Issue #10, opened 2023-08-22
Review Ansible-Lint skipped rules


kpouget:
In this [PR](https://github.com/openshift-psap/topsail/pull/8/files#diff-88353bc8b6867d29d1dbe1e8cee94a3103457e367248fa07610157d8072901e6R11) we updated the Ansible linter, but also introduce multiple "skipped" rules (to keep the CI green).

These skipped rules should be reviewed, and decided one by one if the rule should be enforced or skipped.

Rules that are skipped:
```
  - 'command-instead-of-module'
  - 'command-instead-of-shell'
  - 'deprecated-local-action'
  - 'key-order[task]'
  - 'jinja[spacing]'
  - 'no-free-form'
  - 'chema[meta]'
  - 'name[missing]'
  - 'var-naming[no-reserved]'
  - 'var-naming[no-role-prefix]'
  - 'var-naming[pattern]'
  - 'yaml[colons]'
  - 'yaml[commas]'
  - 'yaml[comments]'
  - 'yaml[empty-lines]'
  - 'yaml[indentation]'
  - 'yaml[key-duplicates]'
  - 'yaml[line-length]'
  - 'yaml[new-line-at-end-of-file]'
  - 'yaml[octal-values]'
  - 'yaml[trailing-spaces]'
  - 'yaml[truthy]'
  - 'name[template]'
  - 'name[casing]'
  - 'risky-file-permissions'
  - 'risky-shell-pipe'
  - 'ignore-errors'
  - 'no-changed-when'
  - 'fqcn'
  - 'args[module]'
```

rules that should be enforced, at first glance:
```

-    command-instead-of-shell
-    jinja[spacing]
-    yaml[new-line-at-end-of-file]
-    risky-shell-pip
-    name[missing]
```

ccamacho:
I think also running some bashate checks anywhere where bash/command is used will be useful

kpouget:
Me too

kpouget:
Me too

PR #16, opened 2023-08-25
testing: lints partial fixes

ccamacho:
This commit resolves the following lint rules:

- 'key-order[task]'
- 'jinja[spacing]'
- 'name[missing]'
- 'yaml[new-line-at-end-of-file]'
- 'yaml[trailing-spaces]'
- 'risky-shell-pipe'
- 'args[module]'

Also removes the branch constraint to run
the Ansible lint checks from any fork. This
is useful because in each contributor's fork
when pushing specific named branches the checks
will be executed, a future improvement could be
to run all the lints checks from a single environment like `tox -e linters` instead of consuming the GH action.

Partially-solves: #10


ccamacho:
Is there a way we could trigger an e2e job for testing the playbooks?

roles/benchmarking/benchmarking_run_mlperf_ssd/tasks/main.yml
```diff
@@ -7,24 +7,24 @@
   command: oc get nodes -l kubernetes.io/hostname={{ benchmarking_run_mlperf_ssd_node_hostname }} -oname
 
 - name: Ensure that the coco dataset PVC exists
-  command:
-    oc get pvc/{{ benchmarking_run_mlperf_ssd_dataset_pvc_name }}
+  shell: |
+    oc get pvc/{{ benchmarking_run_mlperf_ssd_dataset_pvc_name }} \
```
kpouget, line 11: I'm not sure to understand why you turn this `command` into a `shell`? 🤔 
the way I usually proceed is that if there is nothing requiring bash (`|` or `> file`, most of the time), I use a `command`

roles/benchmarking/benchmarking_run_mlperf_ssd/tasks/main.yml
```diff
@@ -7,24 +7,24 @@
...
 
 - name: Fetch the coco dataset PVC definition (debug)
-  shell:
-    oc get pvc/{{ benchmarking_run_mlperf_ssd_pvc_name }}
-       -n {{ benchmarking_run_mlperf_ssd_namespace }}
-       -oyaml
+  shell: |
+    oc get pvc/{{ benchmarking_run_mlperf_ssd_pvc_name }} \
+       -n {{ benchmarking_run_mlperf_ssd_namespace }} \
+       -oyaml \
```
kpouget, line 18: I'm not sure to understand why you add the pipe in `shell: |`
without the pipe, the block is considered as a single line, so the trailing `\` isn't necessary, that makes the block easier to read from my POV
... but at the same time it may be confusing when you're not 100% clear about the semantic 🙃

roles/cluster/cluster_deploy_ldap/tasks/main.yml
```diff
@@ -144,9 +144,12 @@
   # Workaround until `ocm` supports the --insecure flag
 
   - name: Get the cluster ID
-    shell:
+    shell: |
+      set -o pipefail
       ocm describe cluster "{{ cluster_deploy_ldap_cluster_name }}" --json | jq .id -r
     register: cluster_id_cmd
+    args:
+      executable: /bin/bash
```
kpouget, line 152: not sure why you add this `args.executable`  🤔 To force bash, vs having bash or sh?

roles/cluster/cluster_deploy_operator/tasks/main.yml
```diff
@@ -276,33 +277,32 @@
...
-         > {{ artifact_extra_logs_dir }}/catalog_operator.log
+    shell: |
+      oc logs deployment.apps/catalog-operator \
+        -n openshift-operator-lifecycle-manager \
+        > {{ artifact_extra_logs_dir }}/catalog_operator.log
     failed_when: false
-
   - name: Indicate where the Catalog-operator logs have been saved
     debug: msg="The logs of Catalog Operator have been saved in {{ artifact_extra_logs_dir }}/catalog_operator.log"
-
```
kpouget, line 302: I think these 2 lines shouldn't be removed, all the tasks should be separated by an empty line

roles/cluster/cluster_deploy_operator/tasks/main.yml
```diff
@@ -276,33 +277,32 @@
...
     fail: msg="Failed because the operator could not be installed from the CatalogSource"
 
+  always:
+  - name: Store the YAML of the operator CSV that was installed (debug)
+    shell: |
+      oc get ClusterServiceVersion/{{ operator_csv_name }} \
+         -oyaml \
+          -n "{{ cluster_deploy_operator_namespace }}" \
+          > {{ artifact_extra_logs_dir }}/operator_csv.yml
+  - name: Store the YAML of the subscription
```
kpouget, line 299: could you add an empty line here (it wasn't there before, but while we're at it ;-) )

roles/cluster/cluster_ensure_machineset/tasks/main.yml
```diff
@@ -1,13 +1,16 @@
...
 - name: "Check if the cluster already has a {{ machineset_instance_type }} machineset"
-  shell:
+  shell: |
+    set -o pipefail
     oc get machineset -n openshift-machine-api
     {% if machineset_name | length > 0 %}
       -ojson | jq '.items[] | select(.spec.template.spec.providerSpec.value.instanceType=="{{ machineset_instance_type }}" and .metadata.name=="{{ machineset_name }}") | .metadata.name' -r
     {% else %}
       -o=jsonpath='{.items[?(@.spec.template.spec.providerSpec.value.instanceType=="{{ machineset_instance_type }}")].metadata.name}'
     {% endif %}
```
kpouget, line 10: this one will fail with `shell: |`, as currently written

roles/benchmarking/benchmarking_run_mlperf_ssd/tasks/main.yml
kpouget: looks like a typo, it's the only time this happens 🙃 ;-)

kpouget (commented):
thanks for the effort @ccamacho, really appreciated 🙏🏻 

looks good overall, just 2 things:

* the systematic pipe in `shell: |`, from my POV most of the time it looks better without it. Drawback it that it might be confusing if you don't understand the distinction, and it may be bring unexpected errors if you don't remember the semantic when you type the command, like this single `echo` command
```
shell:
  echo hello
  ls /tmp
```
* the `args.executable: /bin/bash`, I'm not sure to understand what's behind it 🤔 My assumption is that Bash must be the default shell. I don't remember if it's an "implicite" assumption, or if I enforce it somewhere (does Ansible allows that? 🤔 🤔 ), so I would prefer removing it from the few places where it was added, except if there's a strong reason to have it!

roles/benchmarking/benchmarking_run_mlperf_ssd/tasks/main.yml
ccamacho: Done

roles/benchmarking/benchmarking_run_mlperf_ssd/tasks/main.yml
ccamacho: Yeah, multiline blocks with the pipe is how bash interprets code blocks by default, I have seen inconsistencies when using special chars in this type of code blocks.

For instance, this specific example breaks the syntax highlight in my code editor https://imgur.com/a/WFaenpt

roles/cluster/cluster_deploy_ldap/tasks/main.yml
ccamacho: Yeah about this, in my case I use ZSH and Ansible is not able to determine what is the default interpreter so it breaks, if this is intended to run I can remove it everywhere to make the PR easier to review ++

kpouget: hum, strange, I've been using zsh as my main shell forever, and I never hard issues 🤔 
what about all the other `shell:` tasks?
can you show me how Ansible complains?

kpouget:
thanks @ccamacho , looks all good to me 👍🏻 

launching a WatsonX test just to be sure

kpouget:
[test passed for what matters to this PR](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_topsail/16/pull-ci-openshift-psap-topsail-main-watsonx-serving-light/1696071227837255680/artifacts/light/) (the failure is unrelated, it's a WatsonX specific failure that I can't explain yet :/ )

kpouget:
@ccamacho I let you rebase the PR and confirm it's good to merge 👍🏻

kpouget:
thanks @ccamacho , merging !

PR #56, opened 2023-09-29
Watsonx: add ability to simply deploy a model

kpouget:

PR #57, opened 2023-09-29
testing: watsonx-serving: test_e2e: add the ability to control if we want to delete the other models or not

kpouget:


kpouget:
missing commit from https://github.com/openshift-psap/topsail/pull/56
merging.

PR #78, opened 2023-10-13
Watsonx: test on the DGX

kpouget:


kpouget:
test passed on the DGX, merging.

PR #114, opened 2023-11-21
Notebooks: relocate to the `projects` directory

kpouget:

PR #115, opened 2023-11-21
Documentation refactor

ccamacho:
This commit includes new documentation
for newcomers.

### DIFF
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -1,4 +1,0 @@
-Contributing
-============
-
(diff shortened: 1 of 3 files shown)


kpouget:
thanks for the Carlos,
the way to extend topsail will change soon when I merge https://github.com/openshift-psap/topsail/pull/114 and the similar PRs that will come, I think we should delay this PR until the refactoring is done, or at least stabilized

kpouget:
thanks @ccamacho , I merge this PR, I will work on top of it

PR #135, opened 2023-11-30
Include the assets to build a topsail Python package

ccamacho:
This commit includes the resource required for
the creation of a Python package called topsail
with the same features and without introducing
any directory structure change change to the current
interfaces calling diretly the run_toolbox.py script.
This commit includes the scripts and jobs so with any
change to the package version there will be a new
released pypi package.

The list of what this commit intends:
- Create the pypi workflow for releasing topsail as a python package in pypi.
- Including the setuptools dependencies to build a python package.
- Organizing the files so the dependencies can be reached from within the
package (running run_toolbox.py <params> and topsail <params> should be equivalent).


setup.py
```diff
@@ -0,0 +1,85 @@
...
+import os
+from sys import version_info
+
+from setuptools import find_packages, setup
+from topsail.__info__ import version
+
+if version_info < (3, 5):
+    raise RuntimeError(
+        'Python 3.5 or greater is required'
+    )
```
kpouget, line 28: I think 3.9 is required

kpouget:
in the final version, could you also include a symlink from `projects` to `topsail/projects`? 

and this subproject [state-signal-utils](https://github.com/openshift-psap/topsail/tree/6f8b2d1c991a4e29c421948930cea09c64133120/subprojects/state-signal-utils) needs to be part of the Python package, otherwise `kserve` and `notebooks` won't work

kpouget:
closing as outdated

PR #145, opened 2023-12-05
Include projects and subprojects in the package path

ccamacho:
This commit includes the projects and subprojects folders as part of the topsail package path.


kpouget:
@ccamacho ,
> topsail/subprojects/llm-load-test
this one actually belongs to `topsail/llm-load-test/subprojects/llm-load-test`
but when I did the relocation, the CI was failing to build the image because of it

so 1/ we need to make sure that you're not hitting the same issue
2/ if you're not hitting this issue, we can try to relocate it, there should be only one variable to change to update the location
(but right now, I can't remember where the location is specified :#)

requirements.txt
```diff
@@ -1,13 +1,13 @@
...
+ansible
+jinja2
 flake8
 pylint
 awscli
 numpy
 joblib
 
 jsonpath_ng
-state-signals==0.5.2
```
ccamacho, line 13: @kpouget quick question, is there a need for the pins on those packages versions? ansible, jinja2, and state-signals?

kpouget: as long as nothing breaks, not really

.github/workflows/ansible-lint.yml
```diff
@@ -22,7 +22,14 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: Set ANSIBLE_ROLES_PATH
-      run: echo "ANSIBLE_ROLES_PATH=\"$(find ./projects -type d -name roles -print | tr '\n' ':')$(ls roles/* -d | tr '\n' ':')\"" >> $GITHUB_ENV
```
ccamacho, line 25: There is no need to get each role path, we only need the 'roles' folder where they all are, and this path needs to be absolute. Also we should preserve the previous paths.

kpouget:
@ccamacho , the test failed because Prow couldn't build the image, because it couldn't find something around git. Same issue I mentioned above :/ 

also, you can't launch the notebook and kserve test simultaneously (for OCP CI, it's the same test 🙃)

kpouget:
closing as outdated

PR #162, opened 2024-01-08
[kserve] Generate LTS documentation files

kpouget:


kpouget:
kserve test passed, merging

PR #166, opened 2024-01-11
[Notebooks] Validate the LTS payload when generating it

kpouget:


kpouget:
tests passed, merging

PR #171, opened 2024-01-15
WIP Running kserve with local-ci on baremetal

drewrip:


projects/kserve/testing/test.py
```diff
@@ -292,7 +292,7 @@ def cleanup_sutest_ns():
     """
 
     label = config.ci_artifacts.get_config("tests.scale.namespace.label")
-    run.run(f"oc delete ns -l{label}")
+    run.run(f"oc delete ns -l{label} --ignore-not-found=true")
```
kpouget, line 295: you shouldn't have any issue with this one (without `--ignore-not-found`), because searching by label shouldn't match anything 🤔 
or you hit a race condition, where `-l` matched something, but by the time `oc` tried to delete it it didn't exist anymore? 🤔 not sure this can happen ... ?

drewrip: Hmmm, I'm not sure. I opened the PR so I could point the local-ci to a new branch. It looks like you're right. For some reason [the local-ci run pod](https://github.com/openshift-psap/topsail/blob/main/projects/local-ci/roles/local_ci_run/templates/pod.yaml.j2) in a failing state without much explanation. I figure there are probably many reasons this could be since I set this node up manually.

PR #176, opened 2024-01-17
WIP POC regression indicators

drewrip:
Some framework to create a number of different regression indicators, and a placeholder for potentially some future indicator that uses polynomial regression. Hoping to use this to discuss more about the design of the Regression portion @kpouget @ccamacho

### DIFF
--- a/topsail/testing/regression.py
+++ b/topsail/testing/regression.py
@@ -1,10 +1,10 @@
 import os
 import json
 
+from datetime import datetime
+
 import numpy as np
 from functools import reduce
-
-from opensearchpy import OpenSearch
 
 regression_return_codes = {
     "POS_REG": 1,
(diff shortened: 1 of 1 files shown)


topsail/testing/regression.py
```diff
@@ -0,0 +1,168 @@
...
+        mean = np.mean(prev_results)
+        std = np.std(prev_results)
+        z_score = (curr_result - mean) / std
+        if abs(z_score) > self.threshold:
+            if z_score < 0:
+                return -1
+            else:
+                return 1
+        else:
+            return 0
```
kpouget, line 86:
```
        if abs(z_score) <= self.threshold:
            status = 0 # in comment: what 0 means?
        elif z_score < 0:
             status = -1 # in comment: what -1 means?
        else:
              status = 1 # in comment: what -1 means?

        return status
```
maybe an enum could be useful here? 🤔

topsail/testing/regression.py
```diff
@@ -0,0 +1,168 @@
...
+    somewhere in the pipeline
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def regression_test(self, curr_result, prev_results):
+        return 0
+
+
+if __name__ == '__main__':
```
kpouget, line 100: when ready for merging, can you move this out of this file?

topsail/testing/regression.py
kpouget: hum, the enum is already there, at the top :D

topsail/testing/regression.py
```diff
@@ -0,0 +1,168 @@
...
+from functools import reduce
+
+from opensearchpy import OpenSearch
+
+regression_return_codes = {
+    "POS_REG": 1,
+    "NEG_REG": -1,
+    "NONE": 0,
+    "UNDEF": 2,
+}
```
kpouget, line 14: can you turn this into a Python enum, and make sure that the string ends up in the final JSON (not the int)
and if you can use full words instead of abbreviations :)

topsail/testing/regression.py
```diff
@@ -0,0 +1,168 @@
...
+class RegressionIndicator:
+
+    def __init__(
+            self,
+            client,
+            kpi,
+            settings=None,
+            combine_func=None,
+            x_var="_source.@timestamp",
+            y_var="_source.value",
```
kpouget, line 26: I'm not sure to understand how (now where) this is used? 🤔 
`_source....` should be internal OpenSearch, this module should not receive OpenSearch values

topsail/testing/regression.py
drewrip: @kpouget Thanks for the feedback! I actually thought I closed this PR, this was a precursor to [the matbench PR](https://github.com/openshift-psap/matrix-benchmarking/pull/125). This is all old code and has been adapted into matrix-benchmarking :)

PR #186, opened 2024-01-24
Add Regression Indicators

drewrip:
This sets up the notebooks performance test to create and use a RegressionIndicator from [this matrix-benchmarking PR](https://github.com/openshift-psap/matrix-benchmarking/pull/121).

The idea is that we can create a number of different indicators from the base RegressionIndicator that provides a generic and flexible result that we can save and process upstream. I created a simple ZScoreIndicator to demonstrate this, it triggers when the new payload's KPI values are more than 3 standard deviations away from the mean value for that KPI among the LTS results. Hopefully we can use this to create a simple wrapper for datastax/Hunter.

There are still 2 work in progress issues to resolve with this PR:
1. **Missing Settings:** Some of the settings for an individual MatrixEntry are only present in the Matrix itself. In our testing payloads, an example would be `version`. I'm not sure why these settings aren't present in the MatrixEntry itself, but as currently written we can only restrict the settings that are present in the MatrixEntry.

2. **Gathered Results:** As discussed some of the LTS payloads have "gathered" results. I wasn't sure how to turn this off and couldn't flatten the `Payload`s since I couldn't deepcopy them.

### DIFF
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -1,17 +1,39 @@
 import json
 import logging
+import numpy as np
+
+from copy import deepcopy
 
 import matrix_benchmarking.common as common
-
+import matrix_benchmarking.regression as regression
 
 def run():
-    logging.info(f"Received {common.Matrix.count_records()} new entries")
-    for entry in common.Matrix.all_records():
-        pass
 
     logging.info(f"Received {common.LTS_Matrix.count_records()} historic LTS entries")
-    for lts_entry in common.LTS_Matrix.all_records(): pass
+    lts_entries_raw = list(common.LTS_Matrix.all_records())
+    # Temporarily skip the gathered results
+    lts_entries = list(filter(lambda entry: type(entry.results) is not list, lts_entries_raw))
+
+    logging.info(f"Received {common.Matrix.count_records()} new entries")
+    new_entries = list(common.Matrix.all_records())
 
     number_of_failures = 0
+    for entry in common.Matrix.all_records():
+        regression_results_dest = entry.location / "regression.json"
+
+        # Check for regression over the image
+        zscore = regression.ZScoreIndicator(
+            entry,
+            lts_entries,
+            settings_filter={"image": "tensorflow:2023.1"},
+            combine_funcs={"notebook_performance_benchmark_time": np.mean}
+        )
+
+        logging.info(f"Saving the regression results in {regression_results_dest}")
+        regression_results = zscore.analyze()
+        with open(regression_results_dest, "w") as f:
+            json.dump(regression_results, f, indent=4)
+            print("", file=f)
+        number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results))
 
     return number_of_failures
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -16,7 +16,7 @@ def run():
     logging.info(f"Received {common.Matrix.count_records()} new entries")
 
     number_of_failures = 0
-    settings_to_check = ["rhoai_version"]
+    settings_to_check = ["rhoai_version", "ocp_version"]
     for entry in common.Matrix.all_records():
         regression_results_dest = entry.location / "regression.json"
         regression_results: List[models.RegressionResult] = []
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -6,6 +6,7 @@ from copy import deepcopy
 
 import matrix_benchmarking.common as common
 import matrix_benchmarking.regression as regression
+import matrix_benchmarking.regression.zscore as zscore
 
 def run():
 
@@ -41,8 +42,8 @@ def run():
                     lts_entries
                 )
             )
-            zscore = regression.ZScoreIndicator(entry, controlled_lts_entries)
-            regression_results[check_setting] = zscore.analyze()
+            zscore_ind = zscore.ZScoreIndicator(entry, controlled_lts_entries)
+            regression_results[check_setting] = zscore_ind.analyze()
             number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results[check_setting]))
 
         logging.info(f"Saving the regression results in {regression_results_dest}")
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -27,8 +27,7 @@ def run():
             except KeyError:
                 logging.warning(f"Couldn't find {check_setting} setting for entry={entry.location}, skipping...")
                 continue
-            
-            controlled_settings = {"image_name": "pytorch"}
+
 
             controlled_lts_entries = list(
                 filter(
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -16,7 +16,7 @@ def run():
     logging.info(f"Received {common.Matrix.count_records()} new entries")
 
     number_of_failures = 0
-    settings_to_check = ["rhoai_version", "image_name"]
+    settings_to_check = ["rhoai_version"]
     for entry in common.Matrix.all_records():
         regression_results_dest = entry.location / "regression.json"
         regression_results: List[models.RegressionResult] = []
@@ -27,6 +27,8 @@ def run():
             except KeyError:
                 logging.warning(f"Couldn't find {check_setting} setting for entry={entry.location}, skipping...")
                 continue
+            
+            controlled_settings = {"image_name": "pytorch"}
 
             controlled_lts_entries = list(
                 filter(
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -2,7 +2,7 @@ import json
 import logging
 import numpy as np
 
-from copy import deepcopy
+from typing import List
 
 import matrix_benchmarking.common as common
 import matrix_benchmarking.regression as regression
@@ -26,7 +26,7 @@ def run():
     settings_to_check = ["version", "repeat"]
     for entry in common.Matrix.all_records():
         regression_results_dest = entry.location / "regression.json"
-        regression_results = {}
+        regression_results: List[models.RegressionResult] = []
         for check_setting in settings_to_check:
             controlled_settings = entry.get_settings()
 
@@ -43,12 +43,18 @@ def run():
                 )
             )
             zscore_ind = zscore.ZScoreIndicator(entry, controlled_lts_entries)
-            regression_results[check_setting] = zscore_ind.analyze()
-            number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results[check_setting]))
+            results: List[models.RegressionResult] = zscore_ind.analyze()
+            # Add back the setting that we are testing since we handled the filtering manually
+            for result in results:
+                result.setting = check_setting
+                regression_results.append(result)
+
+            number_of_failures += sum([x.status for x in results])
 
         logging.info(f"Saving the regression results in {regression_results_dest}")
         with open(regression_results_dest, "w") as f:
-            json.dump(regression_results, f, indent=4)
+            json_results = [dict(m) for m in regression_results]
+            json.dump(json_results, f, indent=4)
             print("", file=f)
 
     return number_of_failures
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -21,17 +21,13 @@ def run():
         regression_results_dest = entry.location / "regression.json"
         regression_results: List[models.RegressionResult] = []
         for check_setting in settings_to_check:
-            controlled_settings = dict(entry.get_settings())
-            try:
-                controlled_settings.pop(check_setting)
-            except KeyError:
-                logging.warning(f"Couldn't find {check_setting} setting for entry={entry.location}, skipping...")
-                continue
-
+            # This will likely need to be updated with the settings that must be equal between entries
+            # for them to be considered similar enough for comparison against LTS entries
+            controlled_settings = {setting: dict(entry.get_settings())[setting] for setting in ["ocp_version"]}
 
             controlled_lts_entries = list(
                 filter(
-                    lambda x: regression.dict_part_eq(controlled_settings, x.get_settings()),
+                    lambda x: regression.dict_part_eq(controlled_settings, dict(x.get_settings())),
                     lts_entries
                 )
             )
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -11,25 +11,17 @@ import matrix_benchmarking.regression.zscore as zscore
 def run():
 
     logging.info(f"Received {common.LTS_Matrix.count_records()} historic LTS entries")
-    lts_entries = []
-    # Take the first result if they were gathered
-    for entry in common.LTS_Matrix.all_records():
-        if entry.results and type(entry.results) is list:
-            entry.results = entry.results[0]
-            lts_entries.append(entry)
-        else:
-            lts_entries.append(entry)
+    lts_entries = list(common.LTS_Matrix.all_records())
 
     logging.info(f"Received {common.Matrix.count_records()} new entries")
 
     number_of_failures = 0
-    settings_to_check = ["version", "repeat"]
+    settings_to_check = ["rhoai_version", "image_name"]
     for entry in common.Matrix.all_records():
         regression_results_dest = entry.location / "regression.json"
         regression_results: List[models.RegressionResult] = []
         for check_setting in settings_to_check:
-            controlled_settings = entry.get_settings()
-
+            controlled_settings = dict(entry.get_settings())
             try:
                 controlled_settings.pop(check_setting)
             except KeyError:
@@ -42,11 +34,13 @@ def run():
                     lts_entries
                 )
             )
-            zscore_ind = zscore.ZScoreIndicator(entry, controlled_lts_entries)
+            if len(controlled_lts_entries) < 1:
+                logging.warning("No LTS entries left after filtering")
+            zscore_ind = zscore.ZScoreIndicator(entry, controlled_lts_entries, check_setting)
             results: List[models.RegressionResult] = zscore_ind.analyze()
             # Add back the setting that we are testing since we handled the filtering manually
             for result in results:
-                result.setting = check_setting
+                result.metric = check_setting
                 regression_results.append(result)
 
             number_of_failures += sum([x.status for x in results])
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -10,30 +10,35 @@ import matrix_benchmarking.regression as regression
 def run():
 
     logging.info(f"Received {common.LTS_Matrix.count_records()} historic LTS entries")
-    lts_entries_raw = list(common.LTS_Matrix.all_records())
-    # Temporarily skip the gathered results
-    lts_entries = list(filter(lambda entry: type(entry.results) is not list, lts_entries_raw))
+    lts_entries = []
+    for entry in common.LTS_Matrix.all_records():
+        if entry.results and type(entry.results) is list:
+            entry.results = entry.results[0]
+            lts_entries.append(entry)
+        else:
+            lts_entries.append(entry)
 
     logging.info(f"Received {common.Matrix.count_records()} new entries")
-    new_entries = list(common.Matrix.all_records())
 
     number_of_failures = 0
     for entry in common.Matrix.all_records():
         regression_results_dest = entry.location / "regression.json"
-
-        # Check for regression over the image
-        zscore = regression.ZScoreIndicator(
-            entry,
-            lts_entries,
-            settings_filter={"image": "tensorflow:2023.1"},
-            combine_funcs={"notebook_performance_benchmark_time": np.mean}
+        regression_results = []
+        controlled_settings = entry.get_settings()
+        controlled_settings.pop("repeat") # Looking for regressions over the version, repeat is a placeholder
+        controlled_lts_entries = list(
+            filter(
+                lambda x: regression.dict_part_eq(controlled_settings, x.get_settings()),
+                lts_entries
+            )
         )
+        zscore = regression.ZScoreIndicator(entry, controlled_lts_entries)
+        regression_results = zscore.analyze()
+        number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results))
 
         logging.info(f"Saving the regression results in {regression_results_dest}")
-        regression_results = zscore.analyze()
         with open(regression_results_dest, "w") as f:
             json.dump(regression_results, f, indent=4)
             print("", file=f)
-        number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results))
 
     return number_of_failures
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
@@ -11,6 +11,7 @@ def run():
 
     logging.info(f"Received {common.LTS_Matrix.count_records()} historic LTS entries")
     lts_entries = []
+    # Take the first result if they were gathered
     for entry in common.LTS_Matrix.all_records():
         if entry.results and type(entry.results) is list:
             entry.results = entry.results[0]
@@ -21,20 +22,28 @@ def run():
     logging.info(f"Received {common.Matrix.count_records()} new entries")
 
     number_of_failures = 0
+    settings_to_check = ["version", "repeat"]
     for entry in common.Matrix.all_records():
         regression_results_dest = entry.location / "regression.json"
-        regression_results = []
-        controlled_settings = entry.get_settings()
-        controlled_settings.pop("repeat") # Looking for regressions over the version, repeat is a placeholder
-        controlled_lts_entries = list(
-            filter(
-                lambda x: regression.dict_part_eq(controlled_settings, x.get_settings()),
-                lts_entries
+        regression_results = {}
+        for check_setting in settings_to_check:
+            controlled_settings = entry.get_settings()
+
+            try:
+                controlled_settings.pop(check_setting)
+            except KeyError:
+                logging.warning(f"Couldn't find {check_setting} setting for entry={entry.location}, skipping...")
+                continue
+
+            controlled_lts_entries = list(
+                filter(
+                    lambda x: regression.dict_part_eq(controlled_settings, x.get_settings()),
+                    lts_entries
+                )
             )
-        )
-        zscore = regression.ZScoreIndicator(entry, controlled_lts_entries)
-        regression_results = zscore.analyze()
-        number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results))
+            zscore = regression.ZScoreIndicator(entry, controlled_lts_entries)
+            regression_results[check_setting] = zscore.analyze()
+            number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results[check_setting]))
 
         logging.info(f"Saving the regression results in {regression_results_dest}")
         with open(regression_results_dest, "w") as f:
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
@@ -49,19 +49,19 @@ class BenchmarkMeasures(matbench_models.ExclusiveModel):
 
 class Results(matbench_models.ExclusiveModel):
     benchmark_measures: BenchmarkMeasures
-    regression: NotebookPerformanceRegression
+
 
 class NotebookPerformanceKPI(matbench_models.KPI, Settings): pass
 
 
 NotebookPerformanceKPIs = matbench_models.getKPIsModel("NotebookPerformanceKPIs", __name__, kpi.KPIs, NotebookPerformanceKPI)
-NotebookPerformanceRegression = matbench_models.RegressionResult
 
 class Payload(matbench_models.ExclusiveModel):
     schema_name: matbench_models.create_schema_field("rhods-notebooks-perf") = Field(alias="$schema")
     metadata: Metadata
     results: Results
     kpis: Optional[NotebookPerformanceKPIs]
+    regression: Optional[List[matbench_models.RegressionResult]] = Field(default=None)
 
     class Config:
         fields = {'schema_name': '$schema'}
--- a/projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
+++ b/projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
@@ -61,7 +61,7 @@ class Payload(matbench_models.ExclusiveModel):
     metadata: Metadata
     results: Results
     kpis: Optional[NotebookPerformanceKPIs]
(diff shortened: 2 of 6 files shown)


projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,13 +1,39 @@
...
 
     number_of_failures = 0
+    for entry in common.Matrix.all_records():
+        regression_results_dest = entry.location / "regression.json"
+
+        # Check for regression over the image
+        zscore = regression.ZScoreIndicator(
+            entry,
+            lts_entries,
+            settings_filter={"image": "tensorflow:2023.1"},
```
kpouget, line 28: can you try to make this dynamic?
`common.Matrix.settings["image"]` should contain all the available images

kpouget: could you try to do the filtering outside of the `regression.ZScoreIndicator` function ? 🤔 
I think it would be clearer if you could pass only the relevant `lts_entries`

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,13 +1,39 @@
...
+            combine_funcs={"notebook_performance_benchmark_time": np.mean}
```
kpouget, line 29: can you use `notebook_performance_benchmark_min_time`, which already provides an "aggregated" value?
I think `min()` is more relevant than `mean()`, but that's open to discussion
if ever we chose to use `mean`, then we should expose it as a KPI instead of `_min`
(the idea is to expose as a KPI the value we're using for the regression analyses)

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,13 +1,39 @@
...
 
 def run():
-    logging.info(f"Received {common.Matrix.count_records()} new entries")
-    for entry in common.Matrix.all_records(): pass
 
     logging.info(f"Received {common.LTS_Matrix.count_records()} historic LTS entries")
-    for lts_entry in common.LTS_Matrix.all_records(): pass
+    lts_entries_raw = list(common.LTS_Matrix.all_records())
+    # Temporarily skip the gathered results
+    lts_entries = list(filter(lambda entry: type(entry.results) is not list, lts_entries_raw))
```
kpouget, line 15: can you try to take the 1st value of the gathered entries? 
in theory, all the entries in `entry.results` should contain the same result +/- noise, as they have been executed in the same conditions

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,13 +1,39 @@
...
+import matrix_benchmarking.common as common
+import matrix_benchmarking.regression as regression
 
 def run():
-    logging.info(f"Received {common.Matrix.count_records()} new entries")
-    for entry in common.Matrix.all_records(): pass
 
     logging.info(f"Received {common.LTS_Matrix.count_records()} historic LTS entries")
-    for lts_entry in common.LTS_Matrix.all_records(): pass
+    lts_entries_raw = list(common.LTS_Matrix.all_records())
```
kpouget, line 13: you don't really need the `list()` here. Without it,  `lst_entries_row` is a generator, that is processed (only once) at the line below
removing `list()` saves memory and processing time, especially if the list gets large

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
drewrip: I should have explained this better. The `combine_funcs={"notebook_performance_benchmark_time": np.mean}` is just an optional way of handling KPIs that provide either a `list[int]` or `list[float]`. The `combine_funcs` option can be omitted and any KPIs that use a list of values will be skipped. All the other KPIs such as `notebook_performance_benchmark_min_time` are analyzed as normal.

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
drewrip: Fair point!

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
drewrip: Sure, I'll add the filtering outside first. It should be as simple as removing the `settings_filter` option. How do you mean dynamic though? As in, we want to run a regression analysis for each unique "image"?

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/184/pull-ci-openshift-psap-topsail-main-rhoai-light/1750426742431944704/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

kpouget:
@drewrip something went wrong with build (matrixbenchmarking commit wasn't available)
try `git commit --amend --no-edit` and force push it, and try to launch the test again

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/184/pull-ci-openshift-psap-topsail-main-rhoai-light/1750426742431944704/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/184/pull-ci-openshift-psap-topsail-main-rhoai-light/1750426742431944704/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/184/pull-ci-openshift-psap-topsail-main-rhoai-light/1750426742431944704/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

drewrip:
[results](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/186/pull-ci-openshift-psap-topsail-main-rhoai-plot/1750972643240054784/artifacts/plot/test/artifacts/4_matbench_analyze_lts.log)

Looks like the analysis took place as expected. No regressions were found but there were a number of issues regrading payload validation. Obviously we need to make sure the payload schemas are consistent, but I'm wondering why the matrix entry location was `/tmp/matrix_benchmarking_results/expe/from_url`.

kpouget:
> [results](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/186/pull-ci-openshift-psap-topsail-main-rhoai-plot/1750972643240054784/artifacts/plot/test/artifacts/4_matbench_analyze_lts.log)
> 
> Looks like the analysis took place as expected. No regressions were found but there were a number of issues regrading payload validation. Obviously we need to make sure the payload schemas are consistent, but I'm wondering why the matrix entry location was `/tmp/matrix_benchmarking_results/expe/from_url`.

good :) 
yes, we'll need to tidy up the entries on OpenSearch!
>  I'm wondering why the matrix entry location was `/tmp/matrix_benchmarking_results/expe/from_url`.
this location is where the entry is downloaded, when launching a "plotting" job, as you did. So this part is expected :)

We'll need to find a way to save the `regression.json` files, even when doing a replot. One way to do it is to include it in the `LTS Documentation` report, around this place: https://github.com/openshift-psap/topsail/blob/main/projects/notebooks/visualizations/rhods-notebooks-performance/plotting/lts_documentation.py#L73
if you do something like that:
```
html.Code(yaml.dump(entry.results.lts.regression_results), style={"white-space": "pre-wrap"})
```
it should appear in the HTML document

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/184/pull-ci-openshift-psap-topsail-main-rhoai-light/1750426742431944704/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/184/pull-ci-openshift-psap-topsail-main-rhoai-light/1750426742431944704/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

projects/notebooks/visualizations/rhods-notebooks-performance/plotting/lts_documentation.py
```diff
@@ -82,7 +83,12 @@ def generateOneLtsDocumentationReport(entry):
     benchmark_measures += [html.Li([html.B("measures:"), html.Code(lts.results.benchmark_measures.measures)])]
 
     results += [html.Ul(benchmark_measures)]
-
     header += [html.Ul(results)]
 
+    regression = []
+    regression += [html.H2("regression analysis")]
+    regression += [html.Code(yaml.dump(entry.results.lts.regression_results, default_flow_style=False), style={"white-space": "pre-wrap"})]
```
kpouget, line 90:
```suggestion
    regression += [html.H1("regression analysis")]
    regression += [html.Code(yaml.dump(lts.regression_results, default_flow_style=False), style={"white-space": "pre-wrap"})]
```
this adjusts the title size, to make it top level, and takes the `regression_results` field from `lts`, which is the focus of this file

kpouget:
@drewrip , the regression results wasn't included (or was empty) in this document:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/186/pull-ci-openshift-psap-topsail-main-rhoai-plot/1751000197686628352/artifacts/plot/test/artifacts/report_01_report:%20LTS%20Documentation.html

and we'll need to define a model defining the format of the `regression_results` field
I guess it should extend the [KPI model](https://github.com/openshift-psap/matrix-benchmarking/blob/main/matrix_benchmarking/models.py#L81), maybe the [notebooks KPI](https://github.com/openshift-psap/topsail/blob/main/projects/notebooks/visualizations/rhods-notebooks-performance/models/kpi.py#L9) (in the TOPSAIL part), so that it can be easily located in OpenSearch, with Grafana

projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
```diff
@@ -37,15 +37,16 @@ class BenchmarkMeasures(matbench_models.ExclusiveModel):
 
 class Results(matbench_models.ExclusiveModel):
     benchmark_measures: BenchmarkMeasures
+    regression: NotebookPerformanceRegression
 
 NotebookPerformanceKPIs = matbench_models.getKPIsModel("NotebookPerformanceKPIs", __name__, kpi.KPIs, kpi.NotebookPerformanceKPI)
+NotebookPerformanceRegression = matbench_models.RegressionResult
```
kpouget, line 43: this isn't necessary, if you don't modify `matbench_models.RegressionResult`, no need to rename it here

projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
```diff
@@ -37,15 +37,16 @@ class BenchmarkMeasures(matbench_models.ExclusiveModel):
...
 
 class Payload(matbench_models.ExclusiveModel):
     schema_name: matbench_models.create_schema_field("rhods-notebooks-perf") = Field(alias="$schema")
     metadata: Metadata
     results: Results
     kpis: Optional[NotebookPerformanceKPIs]
-    regression_results: Optional[Any]
```
kpouget, line 48: can you let this here, in the `Payload`? in MatrixBenchmarking, overall, the `results` are workload specific, but the `payload/metadata/kpis` have a predefined structure (that can be extended, but the core is always the same) , so the regression results belong there. (I need to [browse these entries](https://github.com/openshift-psap/matrix-benchmarking/blob/main/matrix_benchmarking/upload_lts.py#L163) to upload them as KPIs)

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,26 +1,54 @@
...
+        if entry.results and type(entry.results) is list:
+            entry.results = entry.results[0]
+            lts_entries.append(entry)
+        else:
+            lts_entries.append(entry)
+
     logging.info(f"Received {common.Matrix.count_records()} new entries")
+
+    number_of_failures = 0
+    settings_to_check = ["version", "repeat"]
```
kpouget, line 26: I'm not use to understand exactly what's being checked here, ie how you build the reference data used for the regression checking. We'll need to clarify that, it's important to have that clear

kpouget: also, I don't see clearly where you define the KPI being used for regression. Do you test all of them? 🤔
they won't necessarily have the same direction "of failure", eg a `latency` going higher is a failure, but a `bandwidth` going _lower_ is a failure, so you need to define the regression properties for each of them individually, I guess

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,26 +1,54 @@
...
+
+            controlled_lts_entries = list(
+                filter(
+                    lambda x: regression.dict_part_eq(controlled_settings, x.get_settings()),
+                    lts_entries
+                )
+            )
+            zscore_ind = zscore.ZScoreIndicator(entry, controlled_lts_entries)
+            regression_results[check_setting] = zscore_ind.analyze()
+            number_of_failures += sum(map(lambda x: x["regression"]["status"], regression_results[check_setting]))
```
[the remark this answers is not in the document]
kpouget: and, I'm not sure you can `sum` the statuses, if you have `[-1, 1]`, you end up with 0 failures, no?

projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
drewrip: Gotcha, I'll move them under the `Payload`

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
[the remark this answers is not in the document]
drewrip: I need to start moving from functional to Pythonic with `dict` and `list` comprehensions, I'll convert that.

> and, I'm not sure you can sum the statuses, if you have [-1, 1], you end up with 0 failures, no?

In the current way its implemented this is fine. The statuses are really just boolean flags that are set to 1 if there is some anomaly detected. A simple `sum` wouldn't be possible if we would like a more complex status, but more details regarding _what_ the regression looks like, and why it is being reported is contained in the `RegressionResult` structure.

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
drewrip: > I'm not use to understand exactly what's being checked here, ie how you build the reference data used for the regression checking. We'll need to clarify that, it's important to have that clear

I can try to make that clear. Or we can add a note once we finalize it. At the moment it is considering all of the LTS datapoints with the same settings (minus the image version) to be the historical data. The `settings_to_check` is just the settings that we want to check for regression against. We iterate over them to perform the regression testing against each.

> also, I don't see clearly where you define the KPI being used for regression. Do you test all of them? 🤔
they won't necessarily have the same direction "of failure", eg a latency going higher is a failure, but a bandwidth going lower is a failure, so you need to define the regression properties for each of them individually, I guess

Each indicator has an optional parameter called `kpi`. There you can specify a list of KPI's that you'd like to check. I suppose this difference in direction could be implemented in each `RegressionIndicator` we create. However, in the basic `ZScoreIndicator` there is no distinction made. There are two reasons in this simple instance:
1. I figured statistically significant variation in either direction was notable. Having a new image version run significantly faster seems like it would be suspicious? It wouldn't be wrong or bad necessarily but it would be good to make sure we know about them.
2. We also track the direction withing the `RegressionResults`, we can still recover that information if we'd like by filtering out the entries that exhibit regression in the correct direction

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts//test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

kpouget:
@drewrip your link was broken, don't know how that happened 🙃 

> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts//test/artifacts/002__sutest_notebooks__benchmark_performance/

drewrip:
> @drewrip your link was broken, don't know how that happened 🙃
> 
> > https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts//test/artifacts/002__sutest_notebooks__benchmark_performance/

welp, I too am not sure how that happened :)

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.export=true
/var matbench.lts.opensearch.index=rhoai-notebooks-performance

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,17 +1,60 @@
...
     logging.info(f"Received {common.LTS_Matrix.count_records()} historic LTS entries")
-    for lts_entry in common.LTS_Matrix.all_records(): pass
+    lts_entries = []
+    # Take the first result if they were gathered
+    for entry in common.LTS_Matrix.all_records():
+        if entry.results and type(entry.results) is list:
+            entry.results = entry.results[0]
+            lts_entries.append(entry)
+        else:
+            lts_entries.append(entry)
```
kpouget, line 21: I would prefer something like this:
```suggestion        
        data = extract_data_from_gathered(entry) if entry.is_gathered() \
                      else extract_data_from_entry(entry)
```
in particular, this:
```
entry.results = entry.results[0]
```
is a bad practice [in Python]: you receive an object (from the LTS_Matrix), you should not modify it, or you don't know what the future will become 😆 
In Java, `entry.results` would be private, and the compiler would error out on this line, but nothing like that happens in Python!

example, if there were multiple Regression objects, like there are multiple Plotting classes. Your code executes happily, then mine executes and crashes, because `entry.is_gathered -> True` but `entry.results --> not a list` 🙃 

==> I think you should remove this for loop and just have this:
```
lts_entries = list(common.LTS_Matrix.all_records())
```
and include the handling of gathered entries in matbench `regression` package

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
kpouget: > Having a new image version run significantly faster seems like it would be suspicious? It wouldn't be wrong or bad necessarily but it would be good to make sure we know about them.

this is linked with the `must remain stable` vs `should remain stable` distinction that I mentioned here and there.
* If `must remain stable`, any positive or negative variation should be flagged. The binary is the same (the hardware also), so the performance shouldn't change significantly.
* If `should_remain_stable`, the binary will change, like version bumps, so the new version can be faster, without need to raise a red flag for it. But it can be a flashy green flag :D

projects/notebooks/visualizations/rhods-notebooks-performance/models/lts.py
```diff
@@ -60,7 +61,7 @@ class Payload(matbench_models.ExclusiveModel):
...
+    regression: Optional[List[matbench_models.RegressionResult]] = Field(default=None)
```
kpouget, line 64: can you call the field like the type? either `regression/Regression` or `regression_results/RegressionResult`, for consistency

projects/notebooks/visualizations/rhods-notebooks-performance/store/lts_parser.py
```diff
@@ -79,11 +80,16 @@ def generate_lts_metadata(results, import_settings):
...
     results_lts.benchmark_measures = models.lts.BenchmarkMeasures.parse_obj(results.notebook_benchmark)
 
     return results_lts
 
+def generate_lts_regression(results):
+    results_lts = types.SimpleNamespace()
+    regression_results = [] if not results.regression_results else results.regression_results
+    results_lts = [matbench_models.RegressionResult.parse_obj(r) for r in regression_results]
+
+    return results_lts
```
kpouget, line 92: above, you have this:
```
lts_payload.regression = generate_lts_regression(results)
```

can you call the variable `regression`:

```suggestion
def generate_lts_regression(results):
    return [matbench_models.RegressionResult.parse_obj(r) for r in results.regression_results or []]
```

actually, you don't need the variable XD

what I wanted to say is:
```
...
lts_payload.regression = generate_lts_regression(results)
...

def generate_lts_regression(results):
    regression = types.SimpleNamespace()
    ...
    return regression
```
==> same variable / object attribute --> same name. Makes it easier to grep for where this object is used / defined

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,17 +1,60 @@
...
     number_of_failures = 0
+    settings_to_check = ["version", "repeat"]
+    for entry in common.Matrix.all_records():
+        regression_results_dest = entry.location / "regression.json"
+        regression_results: List[models.RegressionResult] = []
+        for check_setting in settings_to_check:
+            controlled_settings = entry.get_settings()
+
+            try:
+                controlled_settings.pop(check_setting)
```
kpouget, line 34: same as above, you cannot (shouldn't) modify the object given to you, except if you know for sure it's a free copy. if it's not (it's not, see below), or you don't know it
```
    def get_settings(self) -> dict:
        return self.settings.__dict__
```

make a copy yourself
```suggestion
            controlled_settings = dict(entry.get_settings())

            try:
                controlled_settings.pop(check_setting)
```

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
```diff
@@ -1,17 +1,60 @@
...
+            except KeyError:
+                logging.warning(f"Couldn't find {check_setting} setting for entry={entry.location}, skipping...")
+                continue
+
+            controlled_lts_entries = list(
+                filter(
+                    lambda x: regression.dict_part_eq(controlled_settings, x.get_settings()),
+                    lts_entries
+                )
+            )
```
kpouget, line 44: this part would be better inside the regression object. You pass all the LTS entries and the settings to check, and inside the class you do the filtering

drewrip: Either way works for me! This functionality is already inside the regression object, I essentially just copied it here per -> https://github.com/openshift-psap/topsail/pull/186#discussion_r1466479723

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
drewrip: Gotcha! I'll try implementing `deepcopy` on the `MatrixEntry` then I'll provide the functionality to flatten it in the `regression` package. Let me know if you had some other method in mind :)

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
drewrip: Got it! Didn't really occur to me that `self.settings`  was returning the dict like that. I suppose I've been spoiled by Rust style ownership

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
kpouget: oh, sorry, I'm going back and forth, sorry sorry sorry 🤦🏻

projects/notebooks/visualizations/rhods-notebooks-performance/analyze/__init__.py
kpouget: no, don't `deepcopy` the entry, try not to modify it and use it as it is
unless you find any fundamental flaw, but the `gathered` aspect is an important concept from my POV

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance
/var matbench.lts.opensearch.upload=false

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance
/var matbench.lts.opensearch.upload=false

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance
/var matbench.lts.opensearch.upload=false

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance
/var matbench.lts.opensearch.upload=false

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance
/var matbench.lts.opensearch.upload=false

drewrip:
Regression workflow working now -> the results are appearing in the [LTS Documentation report](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/186/pull-ci-openshift-psap-topsail-main-rhoai-plot/1757135050928820224/artifacts/plot/test/artifacts/report_02_report:%20LTS%20Documentation.html).

We needed another `parse` pass.

drewrip:
/test rhoai-plot notebooks https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/195/pull-ci-openshift-psap-topsail-main-rhoai-light/1753099883524395008/artifacts/light/test/artifacts/002__sutest_notebooks__benchmark_performance/
/var matbench.workload=rhods-notebooks-performance
/var matbench.config_file=single_notebook_perf.yaml
/var matbench.lts.opensearch.index=regression-test.rhoai-notebooks-performance
/var matbench.lts.opensearch.upload=false

drewrip:
Example of Regression results can be seen [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/186/pull-ci-openshift-psap-topsail-main-rhoai-plot/1757148758761541632/artifacts/plot/test/artifacts/report_02_report:%20LTS%20Documentation.html)

kpouget:
closing as outdated

PR #218, opened 2024-02-28
[kserve] Add watsonx models to config.yaml and minor fixes

dagrayvid:


projects/kserve/testing/config.yaml
```diff
@@ -125,6 +125,9 @@ ci_presets:
 
   # ---
 
+  4min:
+    tests.e2e.llm_load_test.duration: 240
+
```
npalaska, line 130: I might be missing something but can you please explain why you added 4 min test but afaik we are not using it right?

projects/kserve/testing/models/watsonx-llama-2-70b-chat-hf.yaml
```diff
@@ -7,6 +7,7 @@ serving_runtime:
     extra_env:
       FLASH_ATTENTION: "true"
       MAX_CONCURRENT_REQUESTS: 128
-      NUM_GPUS: 8 # Do we want to shard it across 8 GPUs for watsonx testing?
+      NUM_GPUS: 8
```
npalaska, line 10: Some Yaml failure seems to be happening here 🤔

npalaska: And I guess its fixed

kpouget:
thanks @dagrayvid , I merged the dead code removal in https://github.com/openshift-psap/topsail/pull/219

projects/kserve/testing/config.yaml
kpouget: FYI @npalaska (and @dagrayvid), the settings in this section (`.presets`) don't necessarily have to be used right away. The presets can be passed as parameter when launching the test, eg I may want to add a `20min` preset, and launch it manually every once in a while to ensure that it gives the expected results
```
test rhoai-light kserve 20min
```

PR #231, opened 2024-03-03
Increase the liveness and readiness probe

npalaska:
Increase the liveness and readiness probe to give enough time for some of the larger models to load and be ready.


kpouget:
we can close this one I think @npalaska

kpouget:
I close this one as outdated, we don't want to touch the probes for now

PR #241, opened 2024-03-04
Add new model definitions with some updates and remove deprecated ones

npalaska:


projects/kserve/testing/models/watsonx-mpt-7b-instruct2.yaml
```diff
@@ -5,7 +5,9 @@ serving_runtime:
     resource_request:
       nvidia.com/gpu: 1
     extra_env:
+      DEPLOYMENT_FRAMEWORK: "hf_transformers"
       FLASH_ATTENTION: "false"
       MAX_CONCURRENT_REQUESTS: 128
+      MAX_BATCH_SIZE: 96
```
kpouget, line 11: seems that these two parameters should go to `base-standalone-tgis.yaml`, no ?

npalaska: Yeah I guess if we are setting the MAX_BATCH_SIZE same for all the watsonx models. Will do.

npalaska: Fixed.

PR #251, opened 2024-03-05
[pipelines] DSPv2 Testing

drewrip:


kpouget:
hey @drewrip , I let you go on this one, but feel free to ask for help/advise/review/troubleshooting, anything in fact :)

drewrip:
The iris pipeline integration will take a bit more work. After investigating the issue the pipelines are launched in the following [generic datascience notebooks images](https://quay.io/repository/modh/odh-generic-data-science-notebook?tab=tags). They only contain `kfp==1.8.22`, whereas the provided iris pipeline requires kfp.dsl.Dataset which is only in newer versions. I'm not sure if there are images built with newer kfp versions are we can force a newer version.

kpouget:
> The iris pipeline integration will take a bit more work. After investigating the issue the pipelines are launched in the following [generic datascience notebooks images](https://quay.io/repository/modh/odh-generic-data-science-notebook?tab=tags). They only contain `kfp==1.8.22`, whereas the provided iris pipeline requires kfp.dsl.Dataset which is only in newer versions. I'm not sure if there are images built with newer kfp versions are we can force a newer version.

maybe you can install it manually? that's what I was doing when some packages where missing in the images

projects/local-ci/testing/config.yaml
```diff
@@ -41,6 +41,49 @@ workloads:
...
+      test:
+        command: pipelines test test_ci
+      post_cleanup_ns:
+        command: pipelines test cleanup_sutest_ns
+        alway_run: true
+
+  pipelines-stress:
+    command_group: rhods
+    test_args:
+    - stress
```
kpouget, line 60: @drewrip , you should remove the `test_args: [stress]` field from here, and pass it via the command-line, eg:
```
 --test_args "[e2e_perf,\"10min\"]" 
```
so that you can collapse these three configurations into a single one

projects/pipelines/testing/config.yaml
```diff
@@ -33,6 +33,18 @@ ci_presets:
     clusters.sutest.compute.dedicated: false
     clusters.driver.compute.dedicated: false
 
+  stress:
+    extends: [metal]
+    clusters.sutest.compute.dedicated: false
+    clusters.driver.compute.dedicated: false
```
kpouget, line 39: the `metal` profile can be automatically applied if [you call this function](https://github.com/openshift-psap/topsail/blob/007f9991e53e6482714859037fa1819d31054202/projects/kserve/testing/test.py#L47C5-L47C77) in the `init`:
```
is_metal = config.ci_artifacts.detect_apply_metal_profile(METAL_PROFILE)
```
and `...dedicated: false` should go to the metal profile

same comment for the mixed preset below

projects/pipelines/testing/config.yaml
```diff
@@ -125,9 +141,9 @@ ldap:
 rhods:
   catalog:
     image: brew.registry.redhat.io/rh-osbs/iib
-    tag: 655658
-    channel: stable
-    version: 2.6.0
+    tag: 682686
+    channel: fast
```
kpouget, line 145: not important, but no need to change the channel, the RCs are published in these channels:
> fast,alpha,embedded, stable, stable-2.8, eus-2.8

projects/local-ci/testing/config.yaml
drewrip: Sure!

projects/pipelines/testing/config.yaml
drewrip: Is there a reason to use this:
```
is_metal = config.ci_artifacts.detect_apply_metal_profile(METAL_PROFILE)
```
instead of just extending the `metal` profile?

kpouget: yes! if you extend the `metal` profile, your test won't work well in the cloud. Cloud is great for rehearsal, but also if tomorrow we want to compare the results on AWS

drewrip:
[visualizations fixed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/251/pull-ci-openshift-psap-topsail-main-rhoai-light/1783577302018297856/artifacts/light/test/artifacts/002__plots/report_02_report:_user_execution_overview.html) for the new test format

kpouget:
can you add a call to this:
```
run.run(f"testing/utils/generate_plot_index.py > {env.ARTIFACT_DIR}/reports_index.html", check=False)
```
as in [here](https://github.com/openshift-psap/topsail/blob/main/projects/skeleton/testing/test.py#L129), in the right `finally` block, so that it gets always executed? 

it's very convenient not to have to browse all the directories to find the plots 😅

kpouget:
another thing that you could bring in the visualization is [this plot](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/357/pull-ci-openshift-psap-topsail-main-rhoai-light/1782804982454554624/artifacts/light/test/artifacts/000__test-case_common_scheduling/000__matbenchmarking/expe/000__mode=kueue/002__plots/report_06_report:_worker_nodes_load.html):
image

which comes from there: https://github.com/openshift-psap/topsail/blob/main/projects/scheduler/visualizations/schedulers/plotting/resource_allocation.py#L63

it's not perfect, but it gives a first clue of how busy the worker nodes are.
if ever you want to extend it, mind one thing:
you can't easily add two time series.
if you have this, giving you eg the cpu usage every 5s
```
t0->v0, t5->v5, t10->v10
```
but if in the second time series you have that:
```
t1->v'1, t6->v'6, t11->v'11
```
giving you eg the CPU usage of another nodes, also every 5s, but with a +1s offset (--> this is very likely)
then you can't 
* sum them easily
* plot them on top of one another
* etc
that's why I ended up with subplots, instead of my original idea of having one plot with multiple red/blue layers :/

drewrip:
that looks nice, thanks for sharing! I think I'll probably work on adding that for pipelines outside of this effort :). Assuming that I'm capturing enough information for a thorough analysis already. I guess this is different data than what is in the `sutest cluster cpu-memory usage` report?

projects/rhods/roles/rhods_update_datasciencecluster/tasks/main.yml
```diff
@@ -112,6 +112,28 @@
...
+
+- name: Wait till DataScienceCluster is no longer showing
+  shell: |
+    oc get -oname -f "{{ artifact_extra_logs_dir }}/src/datascience_cluster.yaml" --ignore-not-found
+  when: delete_dsc_cmd.rc != 0
+  register: waiting_for_dsc_rm_cmd
+  until: waiting_for_dsc_rm_cmd.stdout == ""
+  retries: 10
+  delay: 6
+
```
kpouget, line 136: you can remove this, I should have fixed it

projects/core/library/visualize.py
```diff
@@ -218,7 +218,7 @@ def call_upload_lts(step_idx, common_args, common_env_str):
     log_file = env.ARTIFACT_DIR / f"{step_idx}_matbench_upload_lts.log"
 
     env_file = pathlib.Path(".env.generated.yaml")
-    generate_opensearch_config_yaml_env(env_file)
+    #generate_opensearch_config_yaml_env(env_file)
```
kpouget, line 221: you should remove the two comments in this file

drewrip:
[small example passing](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/251/pull-ci-openshift-psap-topsail-main-rhoai-light/1787553354600157184/artifacts/light/test/artifacts/002__plots/reports_index.html)

testing/utils/gather-extra.sh
```diff
@@ -115,79 +114,79 @@ queue ${ARTIFACT_DIR}/infrastructure.yaml $OC get infrastructure/cluster
...
+		echo "WARNING: gathering of audit logs on the master disabled on baremetal platforms."
+		continue
+	fi
+	output_dir="${ARTIFACT_DIR}/audit_logs/$path"
+	mkdir -p "$output_dir"
+
+	# Skip downloading of .terminating and .lock files.
+	oc adm node-logs --role=master --path="$path" |
+		grep -v ".terminating" |
+		grep -v ".lock" |
```
kpouget, line 147: @drewrip , this looks wrong to me, the trailing `\` is necessary
can you revert the changes in this file? if I'm not missing anything, they're only cosmetic, and this files comes from OpenShift CI code, so better stay as close to them as possible, even if that won't change much of anything 🙃

drewrip: Gotcha! Thanks for pointing this out. Should be fixed.

kpouget:
[previous test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/251/pull-ci-openshift-psap-topsail-main-rhoai-light/1789734537303953408/artifacts/light/test/artifacts/reports_index.html) ✅

kpouget:
thanks @drewrip , looks good :)
/approve
/lgtm
/hold <-- merge when you want with /unhold

drewrip:
@kpouget Thanks!

PR #265, opened 2024-03-07
[kserve] Add two concurrency presets and dynamically add some llm load test options

npalaska:


npalaska:
Test passed https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-psap_topsail/265/pull-ci-openshift-psap-topsail-main-rhoai-light/1766141352678002688

kpouget:
@npalaska , can you try to give this link when the test passed --> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/265/pull-ci-openshift-psap-topsail-main-rhoai-light/1766141352678002688/artifacts/light/test/artifacts/ or anything relevant inside it (makes it easier to double check what happened during the test)

kpouget:
```
test rhoai-light kserve quick
```
(`kserve` missing in your test launch)

projects/kserve/testing/config.yaml
```diff
@@ -199,15 +199,26 @@ ci_presets:
...
     matbench.lts.opensearch.index: rhoai-kserve-single-light
 
+  watsonx_single_model_gating_heavy:
+    extends: [watsonx_single_model_gating, gating_rehearsal, use_intlab_minio, gating_concurrency_heavy]
+    tests.e2e.models:
+      - watsonx-codellama-34b-instruct-hf
+      - watsonx-llama-2-70b-chat-hf
+    tests.e2e.llm_load_test.max_input_tokens: 3500
+    tests.e2e.llm_load_test.max_output_tokens: 800
+    tests.e2e.llm_load_test.max_sequence_tokens: 3500
```
dagrayvid, line 216: @npalaska I think we should keep the sequence lengths the same for all tests against a specific model. Otherwise the visualizations will be confusing. It make look like throughput scales worse with concurrency than it otherwise would because we are varying the sequence lengths for these higher concurrency tests.

I think my preferred solution would be to use these long sequence lengths for all tests with the larger models, but that will require changing how we split the large models from the small models in the config file

kpouget:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/265/pull-ci-openshift-psap-topsail-main-rhoai-light/1766474414594985984/artifacts/light/test/artifacts/report_index.html), merging.
Thanks @npalaska for the work :)

projects/kserve/testing/config.yaml
npalaska: @dagrayvid  thanks! That's a good point. Yeah, I think we should do that otherwise there is a clear discrepancy between the data samples chosen at higher concurrency vs lower concurrency for these large models. I can think of two possible solutions at this time:
1. We remove `watsonx_single_model_gating` preset and instead have `small` preset that holds models requiring a single GPU and a `big` preset with models requiring more than 1 GPU.
2. We override the `max_input_tokens` `max_output_tokens` etc for the larger models in the `test_e2e.py` file but that could be little ugly, I am not sure

The benefit of 1st is that we maintain a clear distinction in the config file without requiring change of variables anywhere else in the code but we have to run the tests separately for these larger and smaller models.

The benefit of the second is that we wouldn't have to run tests separately but that means we are changing variables in different places in our code.

PR #286, opened 2024-03-13
[kserve] testing: test: fail if the bare-metal cluster profile isn't found 

kpouget:


kpouget:
tested locally, merging

PR #300, opened 2024-03-18
WIP: Test flan-t5-xxl with custom batching env variables

npalaska:


kpouget:
@npalaska , please add the WIP flag in the title or better in the commit to block the merging

and FYI the longevity test is going for still 24h ...

npalaska:
Closing now as it was raised just to test some env variables.

PR #380, opened 2024-05-03
add mistral-7b-instruct-v02 to tested models

willr3:
This should add mistral-7b-instruct-v0.2 to topsail based on my understanding of the configuration in Topsail but I fully expect to need multiple iterations to get the PR in a proper shape. I do not know how to run topsail to test the changes I'm proposing. How do I test running topsail against my branch and the model I'm trying to add?

I did manually deploy the model following the recommended steps and manually ran the validate_model steps to confirm the model could respond to a standard request.

### DIFF
--- a/projects/kserve/testing/config.yaml
+++ b/projects/kserve/testing/config.yaml
@@ -198,6 +198,7 @@ ci_presets:
     - flan-t5-xxl
     - llama-2-13b-chat-hf
     - llama-2-70b-chat-hf
+    - mistral-7b-instruct-v02
     - mpt-7b-instruct2
     - codellama-34b-instruct-hf
     matbench.lts.horreum.test_name: rhoai-kserve-single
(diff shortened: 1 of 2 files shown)


dagrayvid:
This looks good to me assuming you fix the yaml-lint issue. 

@mcharanrm @npalaska, I'm thinking we can just merge this as is, and it will be tested as part of the next CPT run for the next RC release we test. WDYT?

mcharanrm:
Thanks for the heads up about `"mistral-7b-instruct-v02"` model inclusion in the CPT runs. I think this looks good to me but just remember that it increases the execution time of `"topsail"` pipeline further up by 18-27 mins depending on the size of the model.

willr3:
It looks like this failed because the file *has* a trailing newline rather than because it needed a trailing newline? Pushed an update where I remove the newline

PR #381, opened 2024-05-07
[finetuning] Improvements and experimenting

drewrip:


kpouget:
hey @drewrip , I think the new role should be merged with the old role, WDYT?

PR #386, opened 2024-05-16
[pipelines] add config param for MAX_CONCURRENT_RECONCILES

drewrip:


projects/pipelines/testing/test.py
```diff
@@ -211,6 +212,14 @@ def prepare_cluster():
...
         parallel.delayed(prepare_rhods)
 
+    # Update the MAX_CONCURRENT_RECONCILES
+    max_concurrent_reconciles = config.ci_artifacts.get_config("tests.pipelines.max_concurrent_reconciles")
+    updated_param = {
+        "data": {
+            "MAX_CONCURRENT_RECONCILES": str(max_concurrent_reconciles)
+        }
+    }
+    run.run(f"oc patch cm data-science-pipelines-operator-dspo-parameters -n redhat-ods-applications --patch '{json.dumps(updated_param)}'")
```
kpouget, line 222: @drewrip , are you sure that this config map isn't reconcilated to its default value by the RHODS operator?
(just asking, as it's often the case)

drewrip: I didn't notice that in my small scale test, but maybe I didn't wait long enough for it to be reconciled. I'm not sure though, this didn't cross my mind.

kpouget: I wanted to check in the CI `gather-extra` ... but it wasn't captured 😅 
just double check after >10min. If your config is still there, you're good, they took care of it.
In general, as soon as you modify something that was deploy by an operator (=not a resource you created yourself), you need to make sure that the operator owning the CRD doesn't reconciliate it. 
Either the operator allows the modification, because the devs tooks care of it (should be the case here), or you need to turn off the operator so that it doesn't overwrite your update

drewrip: Gotcha! That's a good point. I'll launch something longer to make sure. Hopefully they'll allow the edit. I wasn't able to figure out how I could change these defaults while deploying the operator itself and this seemed simpler but we'll see if it works :)

drewrip:
[test shows the patch was successful](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/386/pull-ci-openshift-psap-topsail-main-rhoai-light/1791193422883721216/artifacts/light/prepare/build-log.txt)

```
INFO:root:run: oc patch cm data-science-pipelines-operator-dspo-parameters -n redhat-ods-applications --patch '{"data": {"MAX_CONCURRENT_RECONCILES": "25"}}'
configmap/data-science-pipelines-operator-dspo-parameters patched
```

Confirming in a different testbed that it stays around long enough.

drewrip:
@kpouget After ~1 hour `MAX_CONCURRENT_RECONCILES` is still set:
```
MAX_CONCURRENT_RECONCILES:
----
25
```
Should be good to merge from my end

projects/pipelines/testing/config.yaml
```diff
@@ -190,6 +190,7 @@ base_image:
     bucket_name: pipelines-scale-test-bucket
 tests:
   pipelines:
+    max_concurrent_reconciles: 10
```
kpouget, line 193:
```suggestion
    max_concurrent_reconciles: null
```
can you make that null, and then:
```
max_concurrent_reconciles = config.ci_artifacts.get_config("tests.pipelines.max_concurrent_reconciles")
if max_concurrent_reconciles is not None:
  oc patch ...
```
this way, by default, you don't touch RHOAI default value and you keep what *they* prefer

drewrip: Makes sense! Just added that

drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/386/pull-ci-openshift-psap-topsail-main-rhoai-light/1791469982907895808/artifacts/light/prepare/build-log.txt) and the MAX_CONCURRENT reconciles was changed when we provided a value

kpouget:
thanks @drewrip , good to merge 
/lgtm
/approve

PR #387, opened 2024-05-20
[launcher] add topsail-launcher files

drewrip:


kpouget:
thanks Drew, 

I'm not sure where we want to host these filles, maybe in `projects/core/subprojects/launcher`, with a link in the main README?

drewrip:
Since I think these will be a common way to interact with TOPSAIL I thought they might fit in a top level directory, but in either case as long as we add a link in the main README I think it'll be a help. Happy to do either, I just wanted to start the process so we could think about maintaining these all in one place :)

kpouget:
@npalaska @dagrayvid @mcharanrm @ccamacho , what do you think about this PR, and where these files should belong in TOPSAIL ?

dagrayvid:
I haven't used `toolbx` so I may be missing some context, but if these scripts will serve as a wrapper around topsail to help build the container and run the toolbox / tests, then I think it makes more sense as a new top level directory than putting it in projects/core/subprojects/launcher

launcher/README.md
```diff
@@ -0,0 +1,65 @@
...
+------------
+
+* You need to have the [Toolbx package](https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/) installed:
+```
+sudo dnf install toolbox
+```
+
+* You need to have TOPSAIL cloned locally
+```
+git clone https://github.com/openshift-psap/topsail.git
```
kpouget, line 23:
```suggestion
git clone https://github.com/openshift-psap/topsail.git
cd topsail
git submodule update --init
```

drewrip:
@kpouget instructions updated to include getting the submodules. And rebased.

kpouget:
thanks, let's merge it :) 
/approve
/lgtm

PR #388, opened 2024-05-20
[pipelines] save fewer unnecessary artifacts during large tests

drewrip:


projects/pipelines/roles/pipelines_run_kfp_notebook/defaults/main/config.yml
```diff
@@ -38,3 +38,6 @@ pipelines_run_kfp_notebook_capture_artifacts: true
 
 # If True, captures the Prometheus DB of the systems.
 pipelines_run_kfp_notebook_capture_prom_db: false
+
+# Does not change the test, but needed to decide which artifacts to save
+pipelines_run_kfp_notebook_user_count: 0
```
kpouget, line 43: @drewrip , can you change that to something like `capture_extra_artifacts`, so that the role doesn't have an "indirect" parameter?

(maybe you can change it as well in other the role where you have `is_small_test`, IIRC)

projects/pipelines/testing/command_args.yml.j2
```diff
@@ -60,6 +60,7 @@ driver/cluster set_project_annotation/test_toleration:
   project: {{ base_image.namespace }}
 
 pipelines run_kfp_notebook:
+  user_count: {{ tests.pipelines.user_count }}
```
kpouget, line 63: here you change that to:
```suggestion
  capture_extra_artifacts: {% if  tests.pipelines.user_count > 50 %}true{% else %}false{% endif %}
```
or you go one step ahead and define `tests.pipelines.extra_artifacts_threshold: 50`
this way, you can easily define `tests.pipelines.extra_artifacts_threshold: 5000` if you want to force the artifacts capture, or `tests.pipelines.extra_artifacts_threshold: 0` if you want to disable it

overall, best is to move all the constants to the config file ... so that they're not hard-coded constants and can be overwritten. (not an absolute rule though, it would be too heavy-weight to really pass everything this way 🙃 )

projects/pipelines/toolbox/pipelines.py
```diff
@@ -39,6 +39,7 @@ def run_kfp_notebook(self,
                          stop_on_exit=True,
                          capture_artifacts=True,
                          capture_prom_db=False,
+                         user_count=0,
```
kpouget, line 42: default for `capture_extra_artifacts` should be `true`. Assume low-scale by default.

drewrip: Yeah that makes sense. I think having the threshold at all misses the point of the problem we were having anyway. I think I like the `capture_extra_artifacts` flag, since there is really only very specific situations where it should be turned off anyway. I'll switch that for both roles.

kpouget:
@drewrip , RHOAI 2.10rc1 is broken for your use case:
```
{"level":"error","ts":"2024-06-18T20:52:48Z","logger":"opendatahub.controllers.DataScienceCluster","msg":"failed to reconcile workbenches on DataScienceCluster","instance.Name":"default-dsc","error":"buildconfigs.build.openshift.io \"rstudio-server-rhel9\" is forbidden: User \"system:serviceaccount:redhat-ods-operator:redhat-ods-operator-controller-manager\" cannot get resource \"buildconfigs\" in API group \"build.openshift.io\" in the namespace \"redhat-ods-applications\"",
```
this happens when you use [workbenches but no dashboard](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/388/pull-ci-openshift-psap-topsail-main-rhoai-e2e/1803153968407252992/artifacts/e2e/prepare/artifacts/003__rhods__update_datasciencecluster/_python.cmd). Should be fixed with the next RC

drewrip:
> @drewrip , RHOAI 2.10rc1 is broken for your use case:
> 
> ```
> {"level":"error","ts":"2024-06-18T20:52:48Z","logger":"opendatahub.controllers.DataScienceCluster","msg":"failed to reconcile workbenches on DataScienceCluster","instance.Name":"default-dsc","error":"buildconfigs.build.openshift.io \"rstudio-server-rhel9\" is forbidden: User \"system:serviceaccount:redhat-ods-operator:redhat-ods-operator-controller-manager\" cannot get resource \"buildconfigs\" in API group \"build.openshift.io\" in the namespace \"redhat-ods-applications\"",
> ```
> 
> this happens when you use [workbenches but no dashboard](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/388/pull-ci-openshift-psap-topsail-main-rhoai-e2e/1803153968407252992/artifacts/e2e/prepare/artifacts/003__rhods__update_datasciencecluster/_python.cmd). Should be fixed with the next RC

Ah ok, I didn't catch that. I'm slightly surprised, I assume this had something to do with acquiring a cluster since this same test passes on icelake

drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/388/pull-ci-openshift-psap-topsail-main-rhoai-e2e/1805311100867055616/artifacts/e2e/test/artifacts/002__plots/report_02_report:_user_execution_overview.html)

projects/pipelines/testing/test.py
```diff
@@ -215,12 +223,11 @@ def prepare_cluster():
...
-            "data": {
-                "MAX_CONCURRENT_RECONCILES": str(max_concurrent_reconciles)
-            }
-        }
-        run.run(f"oc patch cm data-science-pipelines-operator-dspo-parameters -n redhat-ods-applications --patch '{json.dumps(updated_param)}'")
+        rhoai_down() # Pause RHOAI so we can edit a variable in the DSPO
+        run.run(f"oc set env deployment/data-science-pipelines-operator-controller-manager MAX_CONCURRENT_RECONCILES={str(max_concurrent_reconciles)} -n redhat-ods-applications")
+        # This is less than ideal, but I couldn't find a succinct way to wait until the pods
+        # from this deployment have all succesfully been redeployed
+        time.sleep(30)
```
drewrip, line 230: @kpouget I don't love this solution, but while looking over the deployment and configmap resources I wasn't able to notice a particular condition to wait on

drewrip:
@kpouget This should be good to merge from my side. Let me know if you have any comments :)

projects/pipelines/testing/test.py
kpouget:
```
oc set env deployment/data-science-pipelines-operator-controller-manager
```
you update a deployement here, so you can wait ... for the deployment to be ready!

drewrip: Hmmmm ok, let me test that again tomorrow. When I tested it manually before, I don't think it ever left `Ready`. The deployment just slowly rolled out new pods and took down the old.

kpouget: hum, good point. I'm not very familiar with deployment *update*, but there are some other fields in the status that you can track, like the number not available, number ready

projects/pipelines/roles/pipelines_run_kfp_notebook/tasks/main.yml
```diff
@@ -197,7 +197,7 @@
   - name: Create the secret token
     shell:
       oc create secret generic "{{ secret_token_name }}"
-         "--from-literal=token=$(oc create token '{{ sa_name }}' -n '{{ notebook_namespace }}')"
+         "--from-literal=token=$(oc create token '{{ sa_name }}' -n '{{ notebook_namespace }}' --duration=21600s)"
```
kpouget, line 200: I don't see where this is defined? 🤔 
I wanted to check the meaning of `duration`
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_secret_generic/

kpouget:
looks good @drewrip , we can merge it when you want
/approve
/lgtm
/hold

drewrip:
@kpouget thanks, I just need to test this fix that more gracefully waits for the Pipelines operator pods to be redeployed, then I'll be good to merge

drewrip:
[test passing]: the MCR is getting set. @kpouget Good to merge whenever :)

drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/388/pull-ci-openshift-psap-topsail-main-rhoai-e2e/1807775149420711936/artifacts/e2e/test/artifacts/000__pipelines_run_many/000__local_ci__run_multi/artifacts/ci-pod-0/004__pipelines__capture_state/dspo.desc), good to merge

projects/pipelines/roles/pipelines_run_kfp_notebook/tasks/main.yml
kpouget: I still don't understand 🤔 
It's a parameter of `oc create token`, fine ... but you're calling `oc create secret` 🤔 🤔

projects/pipelines/testing/test.py
```diff
@@ -65,13 +65,17 @@ def wrapper(*args, **kwargs):
...
+    # Pause RHOAI by scaling replicas -> 0
+    if run.run(f'oc get deploy/rhods-operator -n redhat-ods-operator 2>/dev/null', check=False).returncode == 0:
+        run.run("oc patch deployment.apps/rhods-operator -n redhat-ods-operator --type merge -p '{\"spec\": {\"replicas\": 0}}'")
+        run.run("oc wait --for jsonpath='{.spec.replicas}'=0 deployment.apps/rhods-operator -n redhat-ods-operator --timeout=5m")
+
+def rhoai_up():
+    # Resume RHOAI by scaling replicas -> 1
+    if run.run(f'oc get deploy/rhods-operator -n redhat-ods-operator 2>/dev/null', check=False).returncode == 0:
+        run.run("oc patch deployment.apps/rhods-operator -n redhat-ods-operator --type merge -p '{\"spec\": {\"replicas\": 1}}'")
+        run.run("oc wait --for jsonpath='{.status.availableReplicas}'=1 deployment.apps/rhods-operator -n redhat-ods-operator --timeout=5m")
```
kpouget, line 78: this could have been stored in `projects/rhods/library/prepare_rhoai.py`. I'll move it next time I need it :)

kpouget:
thanks @drewrip , I merge this version !

projects/pipelines/roles/pipelines_run_kfp_notebook/tasks/main.yml
drewrip: Sorry I don't understand what the confusion is? The full command is being called is:
```
oc create token '{{ sa_name }}' -n '{{ notebook_namespace }}' --duration=21600s
```
This generates the token that is passed to `oc create secret generic ...`.

kpouget: oh I see, I didn't notice `oc create token` was being called in a `$(....)` 🙃

PR #390, opened 2024-05-21
A Jupyter Notebook to perform Regression Analysis for RHOAI Model Serving

mcharanrm:
### Introduction
A Jupyter Notebook that uses Normal Distribution to determine the limits/threshold for Regression Analysis. The Current test results/measurements should be with in +/- 3 standards deviations from the sample mean. 

The same mean is calculated from the sample space and the sample space have measurements/results of the previous version. 

### Highlights of the PR
1. A Table Graph for each KPI(Throughput, ITL, TTFT, TPOT) that summarizes the regression analysis of each data point/measurement.
2. A Line Graph for each KPI that summarizes the performance over different versions.

### Prerequisites
1. You need a `json` file with the following schema in order to execute the Jupyter Notebook. You should provide Absolute Path of the `json` file in the Notebook.
```
{
	"opensearch_cluster": "<A URL to access OpenSearch Cluster and it should be given without trailing />",
	"opensearch_user": "<Username to be used for authentication with OpenSearch Cluster>",
	"opensearch_user_password": "<Password to be used for authentication with OpenSearch Cluster>",
	"ca_cert_file": "<Cert file for SSL Verification>"
}

```


dagrayvid:
I'm thinking it may make more sense to create a new directory one level up (projects/kserve/visualizations/**notebooks** or something like that) and put the notebook there.

ashishkamra:
> The Current test results/measurements should be with in +/- 3 standards deviations from the sample mean.
Isn't that too lenient? do you have examples of current standard deviation and sample mean numbers for the model serving tests ?

mcharanrm:
I believe it is not lenient because we need exactly +/- 3 standard deviations in Normal Distribution to cover all the measurements/data points in the sample space but how we control the data points in the sample space plays a important factor in the regression analysis.

For example, If you could look the Line Graph that I have attached here, you can see that lines remain flat for all the models except `japanese-llama-2-7b-instruct` and `llama-2-13b-chat-hf`. 

newplot (2)


Flat lines tells us that numbers are distributed in a very short range, that also means that standard deviation is very low or negligible. Now, lets take a look at `codellama-34b-instruct-hf` measurements.
```
[
    618.6, 
    618.3,
    624.8,
    619.3,
    612.7,
    620.9,
    621.3,
    623.6
]
```

```
>>> throughput = '''[
...     618.6, 
...     618.3,
...     624.8,
...     619.3,
...     612.7,
...     620.9,
...     621.3,
...     623.6
... ]'''
>>> throughput = eval(throughput)
>>> 
>>> statistics.stdev(throughput)
3.7244127053805225

>>> statistics.stdev(throughput) * 3     # Basically +/- 12 Token/sec
11.17323811614156
```
For model `codellama-34b-instruct-hf` at concurrency `32`, we expect the latest version's value to fall within +/- 12 Tokens/Sec threshold. Anything otherwise will be marked as regression. 


Now lets take a look at one of the uneven lines that we have here from model  `japanese-llama-2-7b-instruct`. 
```
[
    152.8,
    90.3,
    190.1,
    129.9,
    83.6,
    87.5,
    339.7,
    151.1
]
```

```
>>> throughput = '''[
...     152.8,
...     90.3,
...     190.1,
...     129.9,
...     83.6,
...     87.5,
...     339.7,
...     151.1
... ]'''
>>> 
>>> 
>>> throughput = eval(throughput)
>>> 
>>> statistics.mean(throughput)
153.125
>>> 
>>> statistics.mean(throughput) * 3
459.375

```
 And in here, we expect the new version's value to fall within +/- 500 Tokens/Sec threshold. This is wild side of it. 

But again, like I said, It totally depends on how we control the measurements/data points in sample space. I have attached a table which is generated by the notebook as part of regression analysis. If you could see, for `llama-2-13b-chat-hf` at `1` concurrency believed to have regression even though the stddev is very small, which is (0.4 tokens/sec) and the delta is (2.1 tokens/sec). 

regression-analysis-table

ashishkamra:
Thanks for the explanation @mcharanrm. One more question: can you explain what do you mean by 'present_ver_mean' and 'previous_ver_mean' ?

mcharanrm:
Sure I can and apologies for not explaining it in the first place. 

**present_ver_mean**  -  In other words, the average throughput of the latest version.
**previous_ver_mean** - sample mean, Avg of the sample space where space space is a list which has avg throughput from the previous version.
**std_dev** - standard deviation of the sample space.
**std_dev_[1, 2, 3]** - A %percentage value that shows what chunk of sample space fall within 1st, 2nd and 3rd standard deviation.

**std_dev_[1, 2, 3]_bound** - A boolean value that defines the closure property. Does the latest value fall within the 1st, 2nd, and 3rd standard deviation of the sample space ?
**change** - % change in average throughput w.r.t sample mean
**delta** - difference between average throughput and sample mean

mcharanrm:
> I'm thinking it may make more sense to create a new directory one level up (projects/kserve/visualizations/notebooks or something like that) and put the notebook there.

Seems like ideal choice but I would like to hear @kpouget thoughts on it.

kpouget:
> > I'm thinking it may make more sense to create a new directory one level up (projects/kserve/visualizations/notebooks or something like that) and put the notebook there.
> 
> Seems like ideal choice but I would like to hear @kpouget thoughts on it.

🤔 
I'm not 100% sure.

Either:
```
projects/kserve/visualizations/kserve-llm/notebooks/regression-analysis-using-normal-distribution.ipynb
```
if the notebook is 100% focused on `kserve-llm` metrics (which I doublt, in the medium term at least)
or 
```
projects/kserve/visualizations/notebooks/regression-analysis-using-normal-distribution.ipynb
```
if it's related to `kserve-llm` and `kserve-prom`. Any of the two would work IMO.

I would recommend adding a way to launch this notebook and capture its output.
And better if it works with `rhoai-plot` as it doesn't require a cluster.

This mean:
* adding `jupiter` as a dependency, 
* adding a configuration flag in the `config.yaml` + a preset
* adding a hook in [this function](https://github.com/openshift-psap/topsail/blob/main/projects/kserve/testing/test.py#L189) (or perhaps in [this one](https://github.com/openshift-psap/topsail/blob/6d1b197534e5b8d789d8f2402a9f6448b94daa6b/projects/core/library/visualize.py), if we consider this general purpose)
* running the notebook, saving the artifacts
* adapting the code to connect to OpenSearch & download the artifacts, using the secrets from the vault

mcharanrm:
Yeah, notebook is 100% focused on kserve-llm metrics. So I will go with following path that was suggested. 

> projects/kserve/visualizations/kserve-llm/notebooks/regression-analysis-using-normal-distribution.ipynb

And looking at the other recommendations, I believe that they are leaning towards the activity that is defined in PSAP-1392, "Adding regression analysis to topsail pipeline".

kpouget:
@mcharanrm I think this can be closed as outdated.
Feel free to contribute to the analyze module in MatrixBenchmarking if you see things missing
https://github.com/openshift-psap/matrix-benchmarking/tree/main/matrix_benchmarking/analyze

PR #407, opened 2024-06-13
RHOAI Dashboard Scale testing

smandaRH:

PR #412, opened 2024-06-19
[notebooks] testing: test: call 'cluster wait_fully_awake' at the beginning of the test

kpouget:


kpouget:
Will be tested in https://github.com/openshift-psap/topsail/pull/407. Merging.

PR #438, opened 2024-07-10
[Kserve] Enable local testing

kpouget:


kpouget:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/438/pull-ci-openshift-psap-topsail-main-rhoai-light/1813481466282119168/artifacts/light/test/artifacts/report_index.html).

@dagrayvid can you review this PR and let me know if you see anything you don't want to merge?
you can look at the commit one by one, they're logically spit.
(there's a WIP commit + its revert, that's expected. I'll remove it before merging)

dagrayvid:
Thanks @kpouget. I think all of these are good changes.  

WRT 65080f768fa90d289ee6e89e1cb248da88ec9cdf, will this make the full test run fail if one llm-load-test run fails? I'm not sure we want this in the CPT where one model configuration might fail but we want to test the remaining models (assuming the root cause of the failure will not make all the other models fail as well...).

kpouget:
> WRT [65080f7](https://github.com/openshift-psap/topsail/commit/65080f768fa90d289ee6e89e1cb248da88ec9cdf), will this make the full test run fail if one llm-load-test run fails? I'm not sure we want this in the CPT where one model configuration might fail but we want to test the remaining models (assuming the root cause of the failure will not make all the other models fail as well...).

it depends how the orchestration is written. In general, I prefer when errors are not swallowed (which is the case before this commit, there's 0 successes, but the role returns happily. Makes it easy to miss a functional issue if you don't pay attention to the reports).

For kserve/single-model testing, the role error is handled [there](https://github.com/openshift-psap/topsail/blob/main/projects/kserve/testing/test_e2e.py#L272-L294): `failed` keeps track of the models that failed, and after all the models have been tested, an exception (the last one) is raised. 
==> so the test keeps going after the llm-load-test failure, but the overall test will be a failure

kpouget:
[smoke test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/438/pull-ci-openshift-psap-topsail-main-rhoai-light/1815838177110265856/artifacts/light/test/artifacts/report_index.html). Merging.

PR #457, opened 2024-08-08
[pipelines] RHOAI 2.13 testing

drewrip:


kpouget:
@drewrip can this PR be merged?

drewrip:
@kpouget Yep we can merge this now. I was keeping it open in case we wanted to test rc2 but I don't think its necessary

PR #465, opened 2024-08-22
[fine-tuning] Define new Prom metrics

kpouget:


kpouget:
cannot be CI tested at the moment.
Tested locally, merging.

PR #474, opened 2024-09-03
projects/kserve: removed multiple declarations of the same line

mcharanrm:
It looks like the same line has been declared twice with just few lines apart in the global space. So, this is an attempt to remove the duplicate one.


mcharanrm:
the other debug statement is removed in this commit, b20a6f1

kpouget:
🙏🏻

PR #506, opened 2024-09-22
[notebooks] Honor the 'matbench.lts.generate: false' config flag

kpouget:


kpouget:
test passed well enough, mergin

PR #516, opened 2024-09-24
Fix string interpolation

albertoperdomo2:


albertoperdomo2:
@kpouget cc

kpouget:
thanks @albertoperdomo2 , I noticed that this hadn't been tested, but not that the f-string was missing 🤦🏻 
Merging after gh rebase.

PR #518, opened 2024-09-25
Add RHOAI guard for fine-tuning

albertoperdomo2:
One of the things to precheck before launching a fine-tuning job is to make sure RHOAI is installed in the cluster. This guard does that and exit in case it is not installed.

### DIFF
--- a/projects/fine_tuning/testing/test_finetuning.py
+++ b/projects/fine_tuning/testing/test_finetuning.py
@@ -14,6 +14,7 @@ from collections import defaultdict
 
 from projects.core.library import env, config, run, merge_dicts
 from projects.matrix_benchmarking.library import visualize, matbenchmark
+from projects.rhods.library import prepare_rhoai as prepare_rhoai_mod
 
 import prepare_finetuning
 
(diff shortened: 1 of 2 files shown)


kpouget:
thanks,
can you add another guard, something like this:
```
if run.run(f"oc get dsc -ojsonpath={.items[0].status.installedComponents.trainingoperator}", check=False, capture_stdout=True).stdout != "true":
 <fail>
```
(this call could be stored in the rhoai library, something like `component_is_deployed("trainingoperator")`)

kpouget:
tested locally (with and without a prepared environment)
```
ERROR:root:*** Caught an exception during _run_test_and_visualize: RuntimeError: Training Operator not installed, cluster not prepared for fine-tuning
Traceback (most recent call last):
  File "/var/home/kpouget/openshift/topsail/projects/fine_tuning/testing/test_finetuning.py", line 417, in test
    failed = _run_test_and_visualize()
  File "/var/home/kpouget/openshift/topsail/projects/fine_tuning/testing/test_finetuning.py", line 321, in _run_test_and_visualize
    raise RuntimeError(msg)
RuntimeError: Training Operator not installed, cluster not prepared for fine-tuning
```

merging.

Thanks @albertoperdomo2 !

PR #526, opened 2024-09-27
[llm-load-test] subproject: bump llm-load-test

sjmonson:
Bump llm-load-test to include the openai_plugin refactor and dropping of warmup support.


kpouget:
thanks @sjmonson, merging

PR #533, opened 2024-10-02
[fine_tuning] testing: test: apply the cluster-profiles even outside of the pure bare-metal clusters

kpouget:
Before this PR, the `metal_profiles` field was used to apply cluster-specific CI presets when a given bare-metal cluster was detected (by searching a known node name in the cluster nodes). This worked only for BareMetal cluster infrastructures (and no defined infrastructure).
Running in a bare-metal cluster without a cluster-profile would raise an exception, so that the cluster name/type is always set via the custom profile. The reason for that is that the cluster type is a critical value to include in the test KPIs.

With this PR, the `metal_profiles` are always looked up and applied (if found). But the exception is still raised only if no cluster profile is found in a baremetal cluster.


kpouget:
@albertoperdomo2 , can you let me know if you agree with this PR?

albertoperdomo2:
@kpouget LGTM! 👍🏽

PR #538, opened 2024-10-03
[fine-tuning] Allow downloading models from IBM DMF datalake

kpouget:


kpouget:
@albertoperdomo2 , can you please try this PR and see if it works as expected?
the download is progressing on the DGX, maybe you can launch it on the IBM cluster and confirm that the model is downloaded?

I'll integrate it in the orchestration when I have a bit of time

PR #542, opened 2024-10-07
[core] library: ansible_toolbox: fix the _python.gen.cmd file when printing flags with quotes

kpouget:
…


kpouget:
test passed, merging

PR #534, opened 2024-10-02
[TOPSAIL] Add macOS support for topsail container

albertoperdomo2:
In order to be able to work with the Topsail container in MacOS, the container scripts needed a few adjustments that are effectively introduced by this PR. 

It mimics the `toolbx` behavior (simplistically) by using `podman` to create and enter Topsail environments.

PR #553, opened 2024-10-09
[Launcher] Use default config files instead of sample files

kpouget:
Hello @drewrip @albertoperdomo2 , 

I updated a bit the launcher following @albertoperdomo2 's previous [PR](https://github.com/openshift-psap/topsail/pull/534) to use a `xxx.default` file that contains everything and a `xxx.custom` file, where we can customize values. 

I think that's better than `xxx.sample` -> `xxx`, because with the `.sample` file, it's easy to have issues when a new value is added, but users don't update their `xxx` file. Example: the `has_toolbox` script.


albertoperdomo2:
good addition! 👌🏽

PR #558, opened 2024-10-10
[fine_tuning] Include the disk usage in the Prometheus metrics

kpouget:


kpouget:
test kind of pass. merging. We'll see with larger tests if the disk usage is shown as expected.

PR #576, opened 2024-10-23
[kserve] Update to RHOAI 2.15

mcharanrm:
Updated `tag` and `version` fields in `kserve/config.yaml` file for OpenShift AI 2.15.0 RC1 model serving performance validation. 

I will use the exiting CI presents `"cpt_single_model_gating"` and `"vllm_cpt_single_model_gating"` to deploy LLMs using "TGIS standalone servingruntime" & "vLLM ServingRuntime" when launching e2e CPT tests through topsail from middleware CI.

### DIFF
--- a/projects/kserve/testing/config.yaml
+++ b/projects/kserve/testing/config.yaml
@@ -522,9 +522,9 @@ clusters:
 rhods:
   catalog:
     image: brew.registry.redhat.io/rh-osbs/iib
-    tag: 832635
+    tag: 848124
     channel: fast
-    version: 2.14.0
+    version: 2.15.0
     version_name: rc1
     opendatahub: false
     managed_rhoai: false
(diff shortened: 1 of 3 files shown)


mcharanrm:
Both tests have been completed successfully but it's status is marked as failed because few KPIs didn't pass in the regression analysis.

Topsail performs regression analysis for llm-load-test KPIs as-well-as for resource utilization KPIs. Looking at the regression-analysis report, the number of KPIs didn't pass are very less, (3/507), (1/572) and (5/60), which is insignificant. The KPIs didn't pass are neither appearing from same LLM model nor from the same KPI across different models.

The values recorded in the KPIs that didn't pass the regression analysis seem like potential outliers to me. We should acknowledge it safely and consider that this is not a blocker for 2.15.0 RC1 model-serving performance since there is no trace of potential regression identified.

kpouget:
outdated, closing

PR #581, opened 2024-11-05
WIP: [kserve] Add granite-7b-lab for vllm single model

sjmonson:


sjmonson:
Closing as the granite-7b-lab will be deprecated going forward and #576 adds granite-3.0-instruct support.

PR #582, opened 2024-11-06
[cluster] library: prom: with-queries: load the 'metrics.txt' file from the 'testing_dir'

kpouget:


kpouget:
merging this, the sanity has been tested with https://github.com/openshift-psap/topsail/pull/580
(I didn't verify yet that Prometheus capture `with-queries` is fixed though)

PR #583, opened 2024-11-06
Gate QLoRA preset.

albertoperdomo2:


kpouget:
hey @albertoperdomo2 , this looks quite good ! only the 405b model failed, which isn't unexpected given the size of the model ! 
I'm not exactly sure what happens with the upload when an invalid test is hit. Its LTS payload can't be constructed, so it cannot be uploaded, but I don't remember if that blocks all the uploads or not :/

Long story short, can you disable this 405b model for the time being, until you can validate outside of CPT that it works, then rerun the test over the night/weekend ?

albertoperdomo2:
@kpouget I'll launch it tonight so it does not block the cluster for the day 👍🏽

kpouget:
test failed because I had a stray ray job running. We may want to consider including a Ray cleanup in the code (along with the PyTorchJob cleanup)

kpouget:
I'll relaunch the job

kpouget:
thanks @albertoperdomo2 , merging

PR #586, opened 2024-11-15
[fine_tuning] Introduce different flavors of the visualization module for Ray vs FMS

kpouget:


kpouget:
tests passed ❤️ , merging

PR #589, opened 2024-11-19
[fine_tuning] Include Prometheus for Ilab as a new flavor

kpouget:


kpouget:
merging this. will fix the ilab test later.

PR #590, opened 2024-11-19
[fine_tuning] Test on the instructlab cluster

kpouget:


kpouget:
ok, ilab cannot run on CPU or with a Tesla T4. Merging this version as good enough.

PR #600, opened 2024-11-28
[pipelines] testing: config: update to RHOAI 2.16.0 rc2

kpouget:


kpouget:
@dagrayvid, can you have a look at this PR and let me know if you agree?
I updated a bit what you did to deploy RHOAI 2.16, basically to move the brew vs quay distinction inside the `_setup_brew_registry`, instead of having it in all of the components
We'll forget about it, so better have it in a single place, instead of spread everywhere!

PR #621, opened 2025-01-03
Clean fms-hf-tuning entrypoint

albertoperdomo2:
As per [this PR](https://github.com/foundation-model-stack/fms-hf-tuning/commit/0888ce2b39946c9c8b1a93f0a086156bb8bc016b), the `launch_training.py` script was removed and all trainings are launched by calling `sft_trainer` module directly.

### DIFF
--- a/projects/fine_tuning/toolbox/fine_tuning_run_fine_tuning_job/files/entrypoint/run_fms.sh
+++ b/projects/fine_tuning/toolbox/fine_tuning_run_fine_tuning_job/files/entrypoint/run_fms.sh
@@ -19,16 +19,3 @@ if [[ $WORLD_SIZE == 1 ]]; then
     time python /app/accelerate_launch.py
     exit 0
 fi
-echo "Running on $WORLD_SIZE machines with $NUM_GPUS GPUs each."
-
-time accelerate launch \
-     --debug \
-     --machine_rank $RANK \
-     --num_machines $WORLD_SIZE \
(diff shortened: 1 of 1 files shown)


projects/fine_tuning/toolbox/fine_tuning_run_fine_tuning_job/files/entrypoint/run_fms.sh
```diff
@@ -19,16 +19,3 @@ if [[ $WORLD_SIZE == 1 ]]; then
     time python /app/accelerate_launch.py
     exit 0
 fi
-echo "Running on $WORLD_SIZE machines with $NUM_GPUS GPUs each."
```
kpouget, line 22: Hey Alberto 👋🏻 

I think you should update the `if` above, to have something like:
```
if [[ $WORLD_SIZE != 1 ]]; then
  echo "Running with a multi-node configuration. This is not supported at the moment, aborting."
  exit 1
fi
```
and move everything from the current `if` branch one level below

kpouget:
want to merge that @albertoperdomo2 ?

albertoperdomo2:
oh, completely left it behind. I'll update the branch and we can merge it 👍🏽 @kpouget

kpouget:
not sure why this test didn't get merge (by the bot) before 🤔 
/lgtm

albertoperdomo2:
@kpouget not sure either 🤔

PR #623, opened 2025-01-07
[jump_ci] testing: use an env pass list instead of skip lists

kpouget:


kpouget:
Test looks good ❤️ , merging.

PR #635, opened 2025-01-15
[fine_tuning] ilab: correct throughput and new ilab image

tosokin:


projects/fine_tuning/toolbox/fine_tuning_run_fine_tuning_job/templates/ilab_base_config.yaml.j2
```diff
@@ -20,5 +20,6 @@ seed: 42
 cpu_offload_optimizer: true
 distributed_training_framework: fsdp
 cpu_offload_params: true
-is_granite: true
+#is_granite: true
```
kpouget, line 23: can you remove this flag?  no need to keep the old ones IMO

projects/fine_tuning/toolbox/fine_tuning_run_fine_tuning_job/templates/fine_tuning_job.yaml.j2
```diff
@@ -101,6 +101,10 @@ spec:
               mountPath: /mnt/entrypoint
             - name: config-volume
               mountPath: /mnt/config
+            - name: shm-volume
+              mountPath: /dev/shm
+            - name: shared-volume
+              mountPath: /mnt/shared
```
kpouget, line 107: what is the purpose of this directory? I don't see it used anywhere in TOSPAIL, so it it hardcoded in ilab? 🤔

projects/fine_tuning/toolbox/fine_tuning_run_fine_tuning_job/templates/fine_tuning_job.yaml.j2
```diff
@@ -124,6 +128,12 @@ spec:
           - name: storage-volume
             persistentVolumeClaim:
               claimName: {{ fine_tuning_run_fine_tuning_job_pvc_name }}
+          - name: shm-volume
+            emptyDir:
+              medium: Memory
+              sizeLimit: 20Gi
```
kpouget, line 134: I think this should be optional/configurable, we can look at it together if you want, so that you understand better how to do it in TOPSAIL?

projects/fine_tuning/testing/config.yaml
```diff
@@ -730,7 +730,8 @@ tests:
       image: quay.io/rhoai/ray:2.35.0-py311-cu121-torch24-fa26
     ilab:
       enabled: false
-      image: registry.redhat.io/rhelai1/instructlab-nvidia-rhel9@sha256:b3dc9af0244aa6b84e6c3ef53e714a316daaefaae67e28de397cd71ee4b2ac7e
+      image: registry.redhat.io/rhelai1/instructlab-nvidia-rhel9@sha256:525ab53de3829cac1a9aabb73194f49e22da8fdcf12a01c56ece961300cdab0d
```
kpouget, line 733: can you also update the default model used by ilab, to used the granite-8b?

projects/fine_tuning/toolbox/fine_tuning_run_fine_tuning_job/templates/fine_tuning_job.yaml.j2
tosokin: Yes, this is a requirement for ilab.

kpouget: can you show me the error that happens without it? 
from my POV it's unexpected that ilab tries to write in `/mnt/shared` if there's no env variable pointing to it.
I would prefer having nothing here, and having something like this in `run_ilab.sh`:
```
export SHARED_DIR=/mnt/output/shared
mkdir -p "$SHARED_DIR"
```
if it's not possible to get this behavior, we need to report it to ilab, and in this case, fine to include it here as a workaround

kpouget:
thanks for the updates @tosokin 🙏🏻 

Only thing I suggest fixing (can be in another PR) is the default value: you currently hard-coded a default of 20Gi, whereas it used to be disabled, so eg the FMS test would run with 20Gi of SHM as well.
I'll give you the step to follow. Should be just one line to add in the config file :D 

@albertoperdomo2, please have a look. We'll need to add the support for the SHM arg in the ray launcher as well

kpouget:
oh, look at that
```
PR_POSITIONAL_ARGS: ilab_awsfine_tuning fms
PR_POSITIONAL_ARG_0: fine_tuning-rhoai-light
PR_POSITIONAL_ARG_1: ilab_awsfine_tuning
PR_POSITIONAL_ARG_2: fms
```
the PR [parser](https://github.com/openshift-psap/topsail/blob/main/testing/utils/openshift_ci/pr_args.sh) got it all wrong XD

@MML-coder, FYI:
Tanya launched two tests at the same time. Was a mistake, no worries.
* OpenShift CI launched only 1 test (cannot launch two identical tests (2x `rhoai-light`) simultaneously)
* TOPSAIL PR parser wasn't expecting this at all, and it generated the `variables_overrides.yaml` [file](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/635/pull-ci-openshift-psap-topsail-main-rhoai-light/1881262535051579392/artifacts/light/prepare/artifacts/variable_overrides.yaml) (above) all wrong
This parser will have to be re-written from Bash to Python, and this case should be better handled.
A CI cluster was lost here, so would have been better that TOPSAIL picked up the first, or last configuration

Thinking about that, one thing that could be interesting would be to integrate more messaging from TOPSAIL to the Github PR (cc @albertoperdomo2, as we worked together on the [notification system](https://github.com/openshift-psap/topsail/tree/main/projects/repo/toolbox/notifications))
Eg, if TOPSAIL creates a `_WARNING` directory with some messages inside, these messages could be pickup by the notification role and displayed in the status message
image

See this [old file](https://github.com/openshift-psap/topsail/blob/main/testing/utils/logging.sh), which already does something in this mindset.

PR #639, opened 2025-01-17
[Dockerfile] Launcher remove python9 dependency

MML-coder:
While running the topsail launcher locally in the podman https://github.com/openshift-psap/topsail/tree/main/launcher - found that python9 is not needed so removing it build the TOPSAIL image

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -7,13 +7,11 @@ LABEL 	io.k8s.display-name="OpenShift PSAP topsail" \
  	name="topsail" \
 	url="https://github.com/openshift-psap/"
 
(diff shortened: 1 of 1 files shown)


kpouget:
thanks @MML-coder , the two images (full & lightweight, used only to run the the Jump-CI) build correctly,
and the full image [installs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/639/pull-ci-openshift-psap-topsail-main-rhoai-images/1880313795839004672/artifacts/build-logs/topsail-amd64.log) Python 3.9.21, as it was pinned before.

Looks all good, merging this PR.

PR #640, opened 2025-01-20
[fine_tuning] testing: config: remove fms as 'default' workload

kpouget:
Less confusing this way


kpouget:
@albertoperdomo2 FYI I'll merge that soon. FMS isn't the "default" anymore, need to apply the `fms` prefix to run the FMS flavor. I updated multiple presets that were relying on FMS being the default, but maybe I didn't catch all of them. Anyway, at some point you should make a cleanup based on what's outdated, what names should be changed, what should be touched but maybe renamed, etc

PR #647, opened 2025-01-21
[fine_tuning] ilab: make more resilient

kpouget:


kpouget:
@tosokin FYI
* the `project` -> `ns` is because the `oc get projects` command easily breaks when the cluster is downdown
* the `if progress` is to avoid [this failure](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/646/pull-ci-openshift-psap-topsail-main-jump-ci/1881697505335316480/artifacts/ci/005-test/artifacts/test-artifacts-001/005__plots/0_matbench_parse.log), when the test couldn't complete:
```
  File "/opt/topsail/src/projects/fine_tuning/visualizations/ilab_training/store/parsers.py", line 305, in _parse_ilab_logs
    first_step_timestamp = datetime.datetime.fromisoformat(ilab_metrics.progress[0].timestamp)
IndexError: list index out of range
```

PR #659, opened 2025-01-28
[WIP] [fine_tuning]: Reorder Ray tasks

albertoperdomo2:
As per an internal discussion, we want to: `run -> capture -> cleanup` rather than `cleanup -> run -> capture`. Also, we want to capture KubeRay logs.


projects/fine_tuning/toolbox/fine_tuning_ray_fine_tuning_job/tasks/main.yml
```diff
@@ -468,6 +468,14 @@
     ignore_errors: true
     when: fine_tuning_ray_fine_tuning_job_capture_artifacts | bool
 
+  - name: Capture the logs of KubeRay
```
kpouget, line 471: @albertoperdomo2 , could you add the missing space on line 295 please?

PR #660, opened 2025-01-28
Fix use of Ansible's restricted words

albertoperdomo2:
Fix the use of Ansible restricted words in both KServe and Pipelines project.


albertoperdomo2:
cc: @kpouget

kpouget:
nice, thanks @albertoperdomo2 🙏🏻

PR #671, opened 2025-02-10
kserve: patch predictor pod load time

mcharanrm:
`predictor_pod.load_time = condition_times["Ready"] - condition_times["Initialized"]`

The time delta may include the container image download time if the image has not already been downloaded or is not present on the node where the pod is scheduled. As a result, the `predictor_pod.load_time` can vary depending on the image's availability on the node and this metric raised a regression (false-positive) showing that the model_load_duration for llama-3-70b-instruct in 2.17.0 is 453 seconds, whereas in 2.16.0, it was 222 seconds. 

After looking logs, predictor-pod in 2.17.0 has took about the same time (209 sec) to load the model files from local FS to GPUs and with the remaining time spent downloading the vLLM container image from image registry.

To prevent this variability, use `status.containerStatuses[].state.running.startedAt` as the starting timestamp when calculating the model loading time.

[ container-logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/668/pull-ci-openshift-psap-topsail-main-jump-ci/1887598036440322048/artifacts/ci/005-test/artifacts/test-artifacts/000__local_ci__run_multi_e2e_perf_test/artifacts/ci-pod-0/005__llama-3-70b-instruct/002__kserve__deploy_model/artifacts/predictor-pod.log) and [predictor-pod-status.yaml](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_topsail/668/pull-ci-openshift-psap-topsail-main-jump-ci/1887598036440322048/artifacts/ci/005-test/artifacts/test-artifacts/000__local_ci__run_multi_e2e_perf_test/artifacts/ci-pod-0/005__llama-3-70b-instruct/002__kserve__deploy_model/artifacts/predictor-pod.yaml)


mcharanrm:
>please include a condition to fail safe when the container didn't start
I wanted to but I didn’t have the full picture. I wanted to get an opinion before adding a block for exception handling/fail safe

mcharanrm:
>eg, set the load time to 0 if "kserve-container" not in containers_start_time

I used your example as is, as I couldn't come up with a better approach. PTAL

kpouget:
> > eg, set the load time to 0 if "kserve-container" not in containers_start_time
> 
> I used your example as is, as I couldn't come up with a better approach. PTAL

yes, looks safe enough. Overall it's hard to catch such cases, and to think in advance "this value might be missing", so I often "fix the crash" rather than solve it in advance. But this one is easy enough :)
thanks for the fix!

kpouget:
thanks @mcharanrm, LGTM, but please CI test it before merging with the smoke test

PR #678, opened 2025-02-18
[ matrix-benchmarking] Read `JOB_SPEC` from `ansible_env` object instead

mcharanrm:
Executing the `'upload_lts'` command from my personal workstation raised an exception while attempting to locate the `"JOB_SPEC"` variable in the local environment. Looking into the code, I found that the `"ansible_env"` object contains the required data for this variable. The code has now been updated to read `"JOB_SPEC"` from the `"ansible_env"` object instead.

`json_loads` is still required, as the JSON properties are enclosed in a string.


kpouget:
thanks @mcharanrm , can you try to launch a test that executes this?
@albertoperdomo2 FYI as you hit this as well

mcharanrm:
The only test I can think of is the KServe smoke test. Do you think of anything else? One thing I’m sure of is that a KServe smoke test isn't necessary to validate this change, as it doesn’t require a GPU instance. Any smoke test that involves these visualizations should do the job.  What test do you recommend here ? and How do I launch it ?

kpouget:
any test that runs on OpenShift CI (including the Jump CI) and export its artifacts will do
```
    if test_config.get("export_artifacts.enabled", False):
        bucket = test_config.get("export_artifacts.bucket")
        path_prefix = test_config.get("export_artifacts.path_prefix")

        if ansible_env.get("OPENSHIFT_CI") == "true":
            job_spec = json.loads(os.environ["JOB_SPEC"])
```
it's not that I suspect it can go wrong, but smoke testing is critical to ensure that things don't break unexpectedly

PR #55, opened 2023-09-28
testing: watsonx-serving: config: rework the e2e preset

kpouget:

PR #681, opened 2025-02-18
Enable Topsail container in MacOS

albertoperdomo2:
Make sure Topsail container can run everything in MacOS.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -24,18 +24,14 @@ ARG CURL_OPTIONS="--silent --location --fail --show-error"
 ARG OCP_CLI_VERSION=latest
 
 # Detect architecture and set appropriate OCP client URL
-RUN ARCH=$(uname -m) && \
-    if [ "$ARCH" = "aarch64" ] || [ "$ARCH" = "arm64" ]; then \
-        OCP_CLI_URL=https://mirror.openshift.com/pub/openshift-v4/aarch64/clients/ocp/${OCP_CLI_VERSION}/openshift-client-linux.tar.gz; \
-    else \
-        OCP_CLI_URL=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OCP_CLI_VERSION}/openshift-client-linux.tar.gz; \
-    fi && \
+RUN ARCH=$(uname -m | sed s/arm64/aarch64/) && \
+    OCP_CLI_URL=https://mirror.openshift.com/pub/openshift-v4/${ARCH}/clients/ocp/${OCP_CLI_VERSION}/openshift-client-linux.tar.gz && \
     curl ${CURL_OPTIONS} ${OCP_CLI_URL} | tar xfz - -C /usr/local/bin oc
 
 # Install Prometheus with architecture detection
 ARG PROMETHEUS_VERSION=2.36.0
-RUN ARCH=$(uname -m) && \
-    if [ "$ARCH" = "aarch64" ] || [ "$ARCH" = "arm64" ]; then \
+RUN ARCH=$(uname -m | sed s/arm64/aarch64/) && \
+    if [ "$ARCH" = "aarch64" ]; then \
         PROMETHEUS_ARCH="arm64"; \
     else \
         PROMETHEUS_ARCH="amd64"; \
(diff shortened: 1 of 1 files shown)


kpouget (commented):
thanks @albertoperdomo2, that's great !

I suggested a minor improvement, 
and please launch a test, anyone, include the skeleton (`/test rhoai-light skeleton`)
to confirm that the CI likes it

then good to merge when you're ready :)

build/Dockerfile
```diff
@@ -21,21 +21,30 @@ RUN dnf install -y epel-release && \
...
+    if [ "$ARCH" = "aarch64" ] || [ "$ARCH" = "arm64" ]; then \
+        PROMETHEUS_ARCH="arm64"; \
+    else \
+        PROMETHEUS_ARCH="amd64"; \
+    fi && \
+    wget --quiet "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.linux-${PROMETHEUS_ARCH}.tar.gz" -O/tmp/prometheus.tar.gz && \
+    tar xf "/tmp/prometheus.tar.gz" -C /tmp && \
+    mv /tmp/prometheus-${PROMETHEUS_VERSION}.linux-${PROMETHEUS_ARCH}/prometheus /usr/local/bin && \
+    mkdir -p /etc/prometheus/ && \
+    mv /tmp/prometheus-${PROMETHEUS_VERSION}.linux-${PROMETHEUS_ARCH}/prometheus.yml /etc/prometheus/prometheus.yml
```
MML-coder, line 47: At line number #55 , can you change to following so that PYTHONPATH not read as literal.   
Make it explicit instead:
ENV PYTHONPATH=${HOME}

MML-coder: sorry I meant we need to append the $PYTHONPATH to be something like following? Alteast I have to do following to workaround the issue.  

ENV PYTHONPATH="/opt/topsail/src:$HOME"

`bash-5.1# echo $PYTHONPATH 
"/Users/memehta/workspace/topsail"
bash-5.1# ./projects/fine_tuning/testing/test.py prepare_ci
Traceback (most recent call last):
  File "/opt/topsail/src/./projects/fine_tuning/testing/test.py", line 16, in <module>
    from projects.core.library import env, config, run, export, common
ModuleNotFoundError: No module named 'projects'
bash-5.1# export PYTHONPATH=$PYTHONPATH:/opt/topsail/src
bash-5.1# ./projects/fine_tuning/testing/test.py prepare_ci`

Another potentially thing we could add  bind the source /etc/hosts to /etc/hosts in the container.

kpouget: @MML-coder , 
```
bash-5.1# echo $PYTHONPATH 
"/Users/memehta/workspace/topsail"
```
this one doesn't work ... because you don't have TOPSAIL at `/Users/memehta/workspace/topsail`
this is specific to the [launcher](https://github.com/openshift-psap/topsail/tree/main/launcher)
same for passing `/etc/hosts` (I think in Linux we have that for free with the [toolbox](https://containertoolbx.org/) [launcher](https://github.com/openshift-psap/topsail/blob/main/launcher/topsail_enter#L32))

build/Dockerfile
```diff
@@ -21,21 +21,26 @@ RUN dnf install -y epel-release && \
...
-  && tar xf "/tmp/prometheus.tar.gz" -C /tmp \
-  && mv /tmp/prometheus-${PROMETHEUS_VERSION}.linux-amd64/prometheus /usr/local/bin \
-  && mkdir -p /etc/prometheus/ \
-  && mv /tmp/prometheus-${PROMETHEUS_VERSION}.linux-amd64/prometheus.yml /etc/prometheus/prometheus.yml
+RUN ARCH=$(uname -m | sed s/arm64/aarch64/) && \
+    if [ "$ARCH" = "aarch64" ]; then \
+        PROMETHEUS_ARCH="arm64"; \
+    else \
+        PROMETHEUS_ARCH="amd64"; \
+    fi && \
```
kpouget, line 38: @albertoperdomo2 I think you can rewrite this one the same way as `oc`

albertoperdomo2:
This works fine on my side now. Hopefully everything is as it needs to be to work in `arm64` systems and we are not missing anything, but in the worst case, this is a step forward to have the container fully working for `arm64`. 😄

kpouget:
thanks @albertoperdomo2 , looks good
/lgtm

kpouget:
(you'll need to rerun the failed test to turn it green, I think. Rerunning it avoids loosing the LGTM label)

PR #682, opened 2025-02-20
[mac_ai] Keep working

kpouget:


kpouget:
this PR only tests the `mac_ai` and `remote` projects, which aren't used by PSAP. Merging.

PR #683, opened 2025-02-20
[mac_ai] Keep working

kpouget:


kpouget:
merging this version as it doesn't affect PSAP tests.

PR #685, opened 2025-02-20
Kserve: RHOAI 2.18.0 RC performance gating

mcharanrm:
The email didn’t include a tag, so I used a custom tag with the valid SHA digest provided. Clients will always pull the image using the SHA digest. If a valid tag is given but the digest is invalid, clients won’t be able to download the image. 

The SHA digest is only considered when both a tag and digest are provided.

### DIFF
--- a/projects/kserve/testing/config.yaml
+++ b/projects/kserve/testing/config.yaml
@@ -294,10 +294,6 @@ ci_presets:
       testing:
         size: small
         max_concurrency: 32
-    - name: mpt-7b-instruct2
-      testing:
-        size: small
-        max_concurrency: 64
     - name: llama-3-70b-instruct
       testing:
         size: large
--- a/projects/kserve/testing/config.yaml
+++ b/projects/kserve/testing/config.yaml
@@ -503,9 +503,9 @@ clusters:
 rhods:
   catalog:
     image: quay.io/rhoai/rhoai-fbc-fragment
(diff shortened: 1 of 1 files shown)


mcharanrm:
Going ahead with a full test, as I believe all the key changes related to the OpenShift AI operator were already released in 2.17.0. So, I don't expect anything unexpected.

/test jump-ci rhoai-4xh100 kserve vllm_cpt_single_model_gating

mcharanrm:
Removed the `mpt-7b-instruct2` model from the kserve model-serving performance gating test for 2.18.0 RC1, as it is quite outdated. Additionally, we're discussing the deployment issues we encountered with this model when using the latest vLLM version - 'V0 LLM engine (v0.7.3.dev291+gd7637aaec)', to understand whether an issue is required.

Re-launching the kserve model-serving gating test to generate the plots and regression analysis report for the Prometheus data, which failed to generate previously due to some errors.

/test jump-ci rhoai-4xh100 kserve vllm_cpt_single_model_gating

mcharanrm:
CI job was failed because of a regression. The KPI 'itl_max' for 'codellama-34b-instruct-hf' showed an regression at a concurrency of 96. However, this appears to be an outlier, as the KPI 'kserve_llm_load_test_itl_p99', which represents the 99th percentile, is consistent with historical results.

Apart from this, it is on par with 2.17 and 2.16.

PR #695, opened 2025-03-05
[mac-ai] Trim down the plots being generated

kpouget:


kpouget:
Tested in https://github.com/openshift-psap/topsail/pull/693
merging

PR #702, opened 2025-03-10
[mac-ai] Add libomp on MacOS

kpouget:


kpouget:
loos good enough, merging

PR #707, opened 2025-03-14
[mac-ai] Make it work on the M4 node

kpouget:

PR #709, opened 2025-03-14
[TOPSAIL] Update to use Python3.11 and latest llm-load-test

kpouget:


kpouget:
llm load test was tested there:
> https://github.com/openshift-psap/topsail/pull/707#issuecomment-2725153745

kpouget:
should be safe enough. Merging.

